PR #50, opened 2013-12-27
Reply notification

tsileo:
Hi,

I love Isso, but the only thing that prevent me to drop Disqus for Isso is the ability for commenters to receive an email notification of reply.

I get a working version with this feature (I use it in production on my blog: http://thomassileo.com), it still need some improvement, I just open the pull request to open the discussion.

What do you think about this feature ?

I added a field in the comment table, is this something that will prevent you to merge this feature ?

Let me know what you think about this,

Thanks!

### DIFF
--- a/isso/css/isso.scss
+++ b/isso/css/isso.scss
@@ -214,5 +214,11 @@ a {
         }
       }
     }
+
+    > .notification-section {
+        display: none;
+        padding-bottom: 10px;
+      }
+
   }
 }
--- a/isso/db/comments.py
+++ b/isso/db/comments.py
@@ -20,7 +20,7 @@ class Comments:
     """
 
     fields = ['tid', 'id', 'parent', 'created', 'modified', 'mode', 'remote_addr',
-              'text', 'author', 'email', 'website', 'likes', 'dislikes', 'voters']
+              'text', 'author', 'email', 'website', 'likes', 'dislikes', 'voters', 'notification']
 
     def __init__(self, db):
 
@@ -30,7 +30,8 @@ class Comments:
             '    tid REFERENCES threads(id), id INTEGER PRIMARY KEY, parent INTEGER,',
             '    created FLOAT NOT NULL, modified FLOAT, mode INTEGER, remote_addr VARCHAR,',
             '    text VARCHAR, author VARCHAR, email VARCHAR, website VARCHAR,',
-            '    likes INTEGER DEFAULT 0, dislikes INTEGER DEFAULT 0, voters BLOB NOT NULL);'])
+            '    likes INTEGER DEFAULT 0, dislikes INTEGER DEFAULT 0, voters BLOB NOT NULL,',
+            '    notification INTEGER);'])
 
     def add(self, uri, c):
         """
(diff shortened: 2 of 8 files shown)


noqqe:
any news?

pozitron57:
:+1: for this feature!

jacmoe:
It would be really great to have this feature because I have several places where I would like Isso to replace Disqus..

But my sites are not really something that my commenters are following like hawks, so they would probably never notice that they got a reply, let alone even remember that they visited my sites :P

This feature would be the one that actually made me ditch Disqus completely.

dertuxmalwieder:
AFAIR other self-hosted comment systems have it, so you *could* already ditch Disqus...

jacmoe:
> AFAIR other self-hosted comment systems have it, so you could already ditch Disqus...

So, *could* I .. let me see: what other self-hosted systems are there that are as good as Isso?  
I have found none.

dertuxmalwieder:
Last time I checked, HashOver (2.0, not "final" yet) seemed to be pretty decent.

jacmoe:
Hashover has a license that I don't agree with, and it seems a bit clunky. Especially the interface, which is important to me. But, yes: that is pretty close :)

varod:
+1

tofof:
Not clear why this isn't more urgently being integrated.

dkbast:
Is there anything we could do to help the process? This feature is currently the only thing holding me back from making the transition to isso.

blatinier:
Just to answer @dertuxmalwieder on his message on optout not being needed (yes it's more than 2 years ago :D) → optout links in emails are a legal obligation in France. I think France is not the only country requesting such a link. This PR should implement both double opt-in & optout to be completly ok I think.

kaushalmodi:
Looking forward to this feature. While I can enable email notifications when people comment on my blog, I'd like people to optionally start/stop notifications when people reply to their comments.

farmdawgnation:
What's the status of this? This has been open for awhile and still unmerged. I'm looking at Isso pretty seriously, but this is something that's missing imo.

posativ:
There is no opt-out, which is a hard requirement before I can consider merging.

zhangnew:
I think add a checkbox let user choose send email for his comment be replied or not is a good idea.

goxofy:
Maybe someone cloud resolve the conficts with this branch？

rawsh:
Bump. This is a super feature

pellenilsson:
I [merged this to current master](https://github.com/pellenilsson/isso/tree/reply_notification) and implemented opt-out using a link in every mail sent. And I made some small additional improvements:

* Fix for a thread problem that popped up, see #395.
* Add the new column to database if needed (by simply adding the column in a `try` statement, maybe there is a nicer way).
* Include a link to the comment itself in the email (as was already done for the admin emails).

pellenilsson:
There is one thing in this implementation that I'm not happy about, though. If you check the "Subscribe to email notification of replies" checkbox when making a *reply*, you will never get any notifications, since you can't really reply to a reply. I can see a number of solutions:

1. Don't show the checkbox at all when writing replies.
2. Checking the checkbox for a reply could mean that author gets notifications for all replies to the same parent comment that he/she replied to.
3. Even though the *Reply* buttons shown after each reply currently has the exact same effect as clicking *Reply* on the parent comment, it could still be arranged so that notifications are sent based on exactly which *Reply* button you pressed.

Any preferences on this?

rawsh:
I feel like it should act the same as the parent comment. Thanks for update, testing now

pellenilsson:
I agree with @rawsh that solution 2 is the most intuitive one. I updated my branch with this behavior.

FiveYellowMice:
I think solution 3 provides the best user experience, though it doesn't seem it can be implemented cleanly with the current reply model.

blatinier:
@pellenilsson I would very much appreciate your contribution for my own instance and thus consider merging it soon. This current PR is a bit too much outdated and requires some updates on the latest discussions.

pellenilsson:
@vincentbernat @blatinier Working on it! I merged with master again (after the Gravatar and Preview additions), but there are some details left to fix. Will create a new PR after that.

blatinier:
That's great news! Thanks for your work in advance

jalcine:
Oh this is awesome!

dertuxmalwieder:
It is!

rawsh:
Thanks!

PR #148, opened 2015-01-03
Vagrant support

richard-fine:
I've created a Vagrantfile to make it easy to get a test environment up and running with the demo page - Vagrant will spin up a VM with node.js, uWSGI, Apache2, and register everything such that you can immediately view the demo page at http://isso-dev.local/ and start trying out code changes.

Doesn't affect the end user but should make it nicer for me to work with :) maybe helpful for other people too?

### DIFF
--- a/isso/demo/index.html
+++ b/isso/demo/index.html
@@ -8,7 +8,7 @@
 		<div id="wrapper" style="width: 900px; text-align: left; margin-left: auto; margin-right: auto;">
 			<h2><a href="index.html">Isso Demo</a></h2>
 			
-			<script	src="isso/js/embed.min.js"></script>
+			<script	src="../js/embed.min.js"></script>
 
 			<section id="isso-thread" data-title="Isso Test"></section>
 		</div>
--- a/isso/demo/index.html
+++ b/isso/demo/index.html
@@ -7,7 +7,8 @@
 	<div id="page" style="text-align:center;">
 		<div id="wrapper" style="width: 900px; text-align: left; margin-left: auto; margin-right: auto;">
 			<h2><a href="index.html">Isso Demo</a></h2>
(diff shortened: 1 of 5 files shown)


isso/demo/index.html
```diff
@@ -7,7 +7,8 @@
 	<div id="page" style="text-align:center;">
 		<div id="wrapper" style="width: 900px; text-align: left; margin-left: auto; margin-right: auto;">
 			<h2><a href="index.html">Isso Demo</a></h2>
-			<script	src="../js/embed.min.js"></script>
+			
+			<script	src="isso/js/embed.min.js"></script>
```
posativ: The demo site is always available on /demo, changing the JavaScript location breaks things.

richard-fine: Ah, so I should instead just host the whole uWSGI app at / ? Easy enough to change...

richard-fine:
Updated the PR with another commit that reverts the change to the demo page and mounts the whole app at the server root.

posativ:
Thanks. I don't use Vagrant and I wonder how long does it take to deploy (= test?) the application. Actually, you don't need to rebuilt to whole JS code (which takes a while). RequireJS works without compilation to a single file (described in the [Installation](https://posativ.org/isso/docs/install/#install-from-source)).

richard-fine:
Once you've got the box (base VM image) downloaded, it takes about 5 minutes to do 'vagrant up' I think. Now that I understand it's possible to run the demo page under Isso's own web server, the need is less pressing, but it's perhaps helpful to have an environment with a full Apache2+uWSGI setup to test or learn from.

I did try following the instructions in the Installation page to use the JS without compiling it, but I couldn't get it to work. I can have another look now that the rest of it is all working.

PR #158, opened 2015-02-01
Catch exception in anonymize function

rolisz:
I just moved my comments from my Wordpress blog to Isso and some of them had some weird things going on in the IP address field. This should fix it.


posativ:
Thanks!

PR #163, opened 2015-02-17
Fix npm uglifyjs missing package error

albohlabs:
uglifyjs has been [unpublished](https://www.npmjs.com/package/uglifyjs). Instead use uglify-js.


posativ:
That means I've been using the wrong package the whole time? Thanks for the update!

PR #165, opened 2015-02-19
Function "curl"'s "reject" param never used

### LINKED ISSUE #159: When commenting, clicking submit causes errors
I posted this issue on the mailing list, but it seems no one is there...

I installed isso on my site recently, but whenever I tried to comment, I
can see the following error in the log:

    ERROR: POST /new
      File "/usr/local/lib/python2.7/dist-packages/isso/__init__.py", line 133, in dispatch
      File "/usr/local/lib/python2.7/dist-packages/isso/views/comments.py", line 84, in dec
        return func(self, env, req, *args, **kwargs)
      File "/usr/local/lib/python2.7/dist-packages/isso/views/__init__.py", line 45, in dec
      File "/usr/local/lib/python2.7/dist-packages/isso/views/comments.py", line 218, in new
        value=self.isso.sign([rv["id"], sha1(rv["text"])]),
      File "/usr/local/lib/python2.7/dist-packages/isso/__init__.py", line 114, in sign
        return self.signer.dumps(obj)
      File "/usr/local/lib/python2.7/dist-packages/itsdangerous.py", line 566, in dumps
        rv = self.make_signer(salt).sign(payload)
      File "/usr/local/lib/python2.7/dist-packages/itsdangerous.py", line 412, in sign
        return value + sep + self.get_signature(value)
      File "/usr/local/lib/python2.7/dist-packages/itsdangerous.py", line 347, in get_signature
        key = self.derive_key()
      File "/usr/local/lib/python2.7/dist-packages/itsdangerous.py", line 334, in derive_key
        self.secret_key).digest()
    TypeError: cannot concatenate 'str' and 'buffer' objects

The comments were still be sent, but no notification will be received,
with neither smtp or stdout notifier.

Does anyone know why this happened?

xuhdev:
The reject param of "curl" is never used and causes issue when xhr.status >= 500.

---

Calling `reject(xhr.body);` would cause "undefined object cannot be called as a function" issue. I'm not sure whether this is the best way to fix it though; but it seems it's the first step that need to be made to fix #159.

### DIFF
--- a/isso/js/app/api.js
+++ b/isso/js/app/api.js
@@ -35,7 +35,7 @@ define(["app/lib/promise", "app/globals"], function(Q, globals) {
         endpoint = endpoint.substring(0, endpoint.length - 1);
     }
 
-    var curl = function(method, url, data, resolve, reject) {
+    var curl = function(method, url, data, resolve) {
 
         var xhr = new XMLHttpRequest();
 
@@ -51,11 +51,7 @@ define(["app/lib/promise", "app/globals"], function(Q, globals) {
                 document.cookie = cookie;
             }
 
-            if (xhr.status >= 500) {
-                reject(xhr.body);
-            } else {
-                resolve({status: xhr.status, body: xhr.responseText});
-            }
+            resolve({status: xhr.status, body: xhr.responseText});
         }
 
         try {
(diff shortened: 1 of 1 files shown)


posativ:
The reject callback should still be there for future (tm) improvements to the UI feedback. The callback is now only called when it is defined.

xuhdev:
@posativ OK, I understand you want to keep it. But the invoking of `curl` does not have enough parameters in the code.

posativ:
Right, I've pushed some code to work around it for now.

PR #170, opened 2015-02-20
Fix mod_wsgi throwing TypeError

jethrogb:
mod_wsgi errors in https://github.com/GrahamDumpleton/mod_wsgi/blob/3.3/mod_wsgi.c#L3137 on headers in unicode. Make the headers explicit binary.

### DIFF
--- a/isso/views/comments.py
+++ b/isso/views/comments.py
@@ -230,8 +230,8 @@ class API(object):
         self.signal("comments.new:finish", thread, rv)
 
         resp = JSON(rv, 202 if rv["mode"] == 2 else 201)
-        resp.headers.add("Set-Cookie", cookie(str(rv["id"])))
-        resp.headers.add("X-Set-Cookie", cookie("isso-%i" % rv["id"]))
+        resp.headers.add(b"Set-Cookie", cookie(str(rv["id"])))
+        resp.headers.add(b"X-Set-Cookie", cookie("isso-%i" % rv["id"]))
         return resp
 
     def view(self, environ, request, id):
@@ -288,8 +288,8 @@ class API(object):
         rv["text"] = self.isso.render(rv["text"])
 
         resp = JSON(rv, 200)
-        resp.headers.add("Set-Cookie", cookie(str(rv["id"])))
-        resp.headers.add("X-Set-Cookie", cookie("isso-%i" % rv["id"]))
+        resp.headers.add(b"Set-Cookie", cookie(str(rv["id"])))
+        resp.headers.add(b"X-Set-Cookie", cookie("isso-%i" % rv["id"]))
         return resp
 
     @xhr
@@ -325,8 +325,8 @@ class API(object):
 
         resp = JSON(rv, 200)
         cookie = functools.partial(dump_cookie, expires=0, max_age=0)
-        resp.headers.add("Set-Cookie", cookie(str(id)))
-        resp.headers.add("X-Set-Cookie", cookie("isso-%i" % id))
+        resp.headers.add(b"Set-Cookie", cookie(str(id)))
+        resp.headers.add(b"X-Set-Cookie", cookie("isso-%i" % id))
         return resp
 
     def moderate(self, environ, request, id, action, key):
--- a/isso/wsgi.py
+++ b/isso/wsgi.py
@@ -133,17 +133,17 @@ class CORSMiddleware(object):
 
         def add_cors_headers(status, headers, exc_info=None):
             headers = Headers(headers)
-            headers.add("Access-Control-Allow-Origin", self.origin(environ))
-            headers.add("Access-Control-Allow-Credentials", "true")
-            headers.add("Access-Control-Allow-Methods", ", ".join(self.methods))
+            headers.add(b"Access-Control-Allow-Origin", self.origin(environ))
+            headers.add(b"Access-Control-Allow-Credentials", "true")
+            headers.add(b"Access-Control-Allow-Methods", ", ".join(self.methods))
             if self.allowed:
-                headers.add("Access-Control-Allow-Headers", ", ".join(self.allowed))
+                headers.add(b"Access-Control-Allow-Headers", ", ".join(self.allowed))
             if self.exposed:
-                headers.add("Access-Control-Expose-Headers", ", ".join(self.exposed))
+                headers.add(b"Access-Control-Expose-Headers", ", ".join(self.exposed))
             return start_response(status, headers.to_list(), exc_info)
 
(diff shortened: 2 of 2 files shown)


jethrogb:
N.B. Travis CI build errors not due to these changes.

jethrogb:
I'm not sure how to make this work with Python 3.

albohlabs:
I don't think this is a problem of isso, cause in the [werkzeug doc of the Headers class](https://github.com/mitsuhiko/werkzeug/blob/master/werkzeug%2Fdatastructures.py#L1082) a string literal is used.

What exactly is the error message?

jethrogb:
"expected byte string object for header name, value of type unicode found"

posativ:
Your change breaks several unittests and also by PEP-3333 the type is not specified. Werkzeug uses unicode, as mentioned.

There are a few posts on StackOverflow, I might need to research a bit here.

jethrogb:
Note that only the Python 3 unit tests are broken. Here's the relevant section from the PEP:

> Again, all objects referred to in this specification as "strings" must be of type str or StringType , and must not be of type unicode or UnicodeType . And, even if a given platform allows for more than 8 bits per character in str / StringType objects, only the lower 8 bits may be used, for any value referred to in this specification as a "string".
>
> For values referred to in this specification as "bytestrings" (i.e., values read from wsgi.input , passed to write() or yielded by the application), the value must be of type bytes under Python 3, and str in earlier versions of Python.

jethrogb:
The issue is that isso does ```from __future__ import unicode_literals``` which turns ```type("test")``` into ```unicode``` while it should be ```str```.

posativ:
You are citing PEP-333, PEP-3333 actually says:

> WSGI therefore defines two kinds of "string":
>
> "Native" strings (which are always implemented using the type named str ) that are used for request/response headers and metadata
> "Bytestrings" (which are implemented using the bytes type in Python 3, and str elsewhere), that are used for the bodies of requests and responses (e.g. POST/PUT input data and HTML page outputs).

jethrogb:
Definitely citing PEP-3333, just a different section. https://www.python.org/dev/peps/pep-3333/#unicode-issues

Also:
```
Python 2.7.6 (default, Mar 22 2014, 22:59:56) 
>>> type("test")
<type 'str'>
>>> type(b"test")
<type 'str'>
>>> from __future__ import unicode_literals
>>> type("test")
<type 'unicode'>
>>> type(b"test")
<type 'str'>
```

```
Python 3.3.2 (default, Jul 29 2013, 12:21:44) 
>>> type("test")
<class 'str'>
>>> type(b"test")
<class 'bytes'>
>>> from __future__ import unicode_literals
>>> type("test")
<class 'str'>
>>> type(b"test")
<class 'bytes'>
```

posativ:
I see the issue, but I think it should by handled by the WSGI framework. I won't merge your PR, because it breaks Python 3 because the native string for Python 3 is "str".

jethrogb:
I agree this shouldn't be merged.

However, this is not an issue with werkzeug. Because you import unicode_literals, you're passing unicode strings to werkzeug in python 2.7 while you should be passing regular strings.

posativ:
`import unicode_literals` only imports the current module, the code doesn't pass a raw list of headers to `start_response(...)` so werkzeug should take care of this.

posativ:
I think it can be easily fixed with a wsgi middleware that converts all unicode headers to native strings on Python 2.

posativ:
What version of werkzeug do you use? In werkzeug 0.10 (just updated) headers are automatically converted to the "native string" (str for Python 2) [1]. Which HTTP request actually breaks in mod_wsgi? There might be a header added without using werkzeug's `Header` datastructure.

[1] https://github.com/mitsuhiko/werkzeug/blob/master/werkzeug/datastructures.py#L1193-L1194

jethrogb:
I'm using 0.8.1.

posativ:
You can use virtualenv to use a more recent version of werkzeug. Werkzeug 0.8 is only supported for Debian Wheezy and I plan to drop it once Jessie is released.

posativ:
Just quickly tested, but here's a patch (for the legacy/0.9 branch):

```diff
diff --git a/isso/__init__.py b/isso/__init__.py
index 0df6af2..01b09fd 100644
--- a/isso/__init__.py
+++ b/isso/__init__.py
@@ -192,7 +192,7 @@ def make_app(conf=None, threading=True, multiprocessing=False, uwsgi=False):
         allowed=("Origin", "Referer", "Content-Type"),
         exposed=("X-Set-Cookie", "Date")))
 
-    wrapper.extend([wsgi.SubURI, ProxyFix])
+    wrapper.extend([wsgi.SubURI, ProxyFix, wsgi.LegacyWerkzeugMiddleware])
 
     return reduce(lambda x, f: f(x), wrapper, isso)
 
diff --git a/isso/wsgi.py b/isso/wsgi.py
index 50214f3..d933c39 100644
--- a/isso/wsgi.py
+++ b/isso/wsgi.py
@@ -2,6 +2,7 @@
 
 from __future__ import unicode_literals
 
+import sys
 import socket
 
 try:
@@ -149,6 +150,25 @@ class CORSMiddleware(object):
         return self.app(environ, add_cors_headers)
 
 
+class LegacyWerkzeugMiddleware(object):
+
+    def __init__(self, app):
+        self.app = app
+
+    def __call__(self, environ, start_response):
+
+        def to_native(x, charset=sys.getdefaultencoding(), errors='strict'):
+            if x is None or isinstance(x, str):
+                return x
+            return x.decode(charset, errors)
+
+        def fix_headers(status, headers, exc_info=None):
+            headers = [(to_native(key), value) for key, value in headers]
+            return start_response(status, headers, exc_info)
+
+        return self.app(environ, fix_headers)
+
+
 class Request(_Request):
 
     # Assuming UTF-8, comments with 65536 characters would consume

```

jethrogb:
Seems to work when applied to tag 0.9.8

jethrogb:
This is still broken. I'm now using werkzeug 0.9.4. Here's what headers looks like after `fix_headers` has been called:
```
<type 'str'> Date
<type 'str'> Etag
<type 'str'> Cache-Control
<type 'str'> Expires
<type 'str'> Content-Type
<type 'str'> Content-Length
<type 'str'> Last-Modified
<type 'unicode'> Access-Control-Allow-Origin
<type 'unicode'> Access-Control-Allow-Credentials
<type 'unicode'> Access-Control-Allow-Methods
<type 'unicode'> Access-Control-Allow-Headers
<type 'unicode'> Access-Control-Expose-Headers
```

jelmer:
Since we've dropped Python 2 support, I'm going to close this.

PR #172, opened 2015-02-25
Save the session-key as a unicode string in the db

xuhdev:
The session-key should be saved as a string, not a byte string.

### DIFF
--- a/isso/db/preferences.py
+++ b/isso/db/preferences.py
@@ -7,7 +7,7 @@ import binascii
 class Preferences:
 
     defaults = [
-        ("session-key", binascii.b2a_hex(os.urandom(24))),
+        ("session-key", binascii.b2a_hex(os.urandom(24)).decode('utf-8')),
     ]
 
     def __init__(self, db):


xuhdev:
@posativ I believe I found the root cause this time. It seems that when creating a new database, what was put into the database for a session-key is actually a byte sequence, not a string. You didn't have this problem because you are using an old database, not newly created ones. This pull request should solve this problem.

`print(binascii.b2a_hex(os.urandom(24)))` gives `b'1d794016f952ef113667f7428335866452e9711f2366acaf'`, which need to be decoded.

Thanks a lot for helping me debug this problem.

xuhdev:
(Sorry for the spamming) In python 2, `binascii.b2a_hex(os.urandom(24))` is enough to produce str, but in python 3, you have to do decoding, which also works for python 2.7. I have this problem because I created the database with the Python3 version.

posativ:
Indeed, thanks for the debugging and the patch!

PR #182, opened 2015-04-02
make salt configurable by data-isso-salt

wendal:
make salt configurable by data-isso-salt


posativ:
Will you add more functionality? It is currently stored to an unused variable.

Issue #193, opened 2015-08-09
Port from httplib to requests


mathstuf:
The problem I'm running into is that isso is getting 301 redirection status codes which makes isso claim "No URL". I tried changing just `utils.http` to use requests, but it looks like the API is just not compatible with httplib, so a larger change will be needed.

mcepl:
Instead of the cargocultic switching to ``requests`` could you please describe what exact error happens to you? Of course, ``httplib`` can deal with 301 redirect correctly, or in whatever way you want.

mathstuf:
It's been a while, but the response was probably straight 301 response instead of doing the redirect follow (I do http -> https via 301 on my server).

mcepl:
Is http://www.diveintopython.net/http_web_services/redirects.html what you are looking for?

mathstuf:
It would seem so.

PR #194, opened 2015-08-09
http: use requests to handle redirects

mathstuf:
Raw use of httplib does not respect the 3xx status codes which may
result in 200 at the end. Use requests to handle any redirects and loop
detection.

---
Looking again, it wasn't so much. Addresses #193.


posativ:
Obsolete, has been resolved without pulling requests.

PR #199, opened 2015-08-25
Add configuration to make email a required field

NickHu:
This patch adds a parameter to make requiring an email address configurable.

Could probably be improved by:

* ~~write tests~~
* using HTML5 form validation
* not using a nasty regex substitute to remove `(required)` from the placeholder (or internationalised equivalent)


posativ:
Thanks for the contribution. Email verification is not easy and usually requires a validation email being sent. I am okay with the current implementation though.

PR #202, opened 2015-09-15
Update Russian translation

pozitron57:
Pluralization will work just fine now with Russian. Also I corrected several typos and added translation for missed "comment-hidden". But to be honest, I installed Isso via pip and the file structure is completely different at my server.


posativ:
Hey,

I cannot verify the russian translation, but thanks for the suggestions. The JavaScript modules are compiled into a single file for distribution, hence the different structure.

pozitron57:
OK, now I see why the structure is different, thanks. 
For the translation —  I'm using it quite for a while and would notice bugs, if any. So I hope it should be fine.

Issue #79, opened 2014-04-20
limit nesting level to 1


posativ:
To ease #15, the internal comment tree should be the same as exposed by the JS UI (nested by max. 1 level). A good read why nesting is bad: http://blog.codinghorror.com/web-discussions-flat-by-design/

This is rather an internal change.

PR #211, opened 2015-10-29
Configurable nesting limit

### LINKED ISSUE #210: Make comment nesting level configurable
The current Isso JavaScript client allows for only one level of comment nesting. While this may be a good UX choice for most, there are many situations where arbitrary levels of comment nesting are desired.

The Isso client should support a user-configurable nesting level limit. The default should be "1", or the same limit as currently imposed, for backwards compatibility.

sprin:
Resolves #210

### DIFF
--- a/docs/docs/configuration/client.rst
+++ b/docs/docs/configuration/client.rst
@@ -100,6 +100,8 @@ data-isso-nesting-level
 Maximum level of comment nesting allowed. The default value of `"1"` will allow
 comments only at the top-level and as direct replies to top-level comments.
 
+Set to `"inf"` to allow unlimited nesting.
+
 data-isso-id
 ------------
 
--- a/docs/docs/configuration/client.rst
+++ b/docs/docs/configuration/client.rst
@@ -94,6 +94,12 @@ data-isso-vote
 
 Enable or disable voting feature on the client side.
 
+data-isso-nesting-level
+--------------
+
+Maximum level of comment nesting allowed. The default value of `"1"` will allow
+comments only at the top-level and as direct replies to top-level comments.
+
 data-isso-id
 ------------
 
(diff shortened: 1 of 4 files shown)


pozitron57:
Let me clarify, does it allow to set any nesting level (including 0 = nesting disabled)?

sprin:
Yes, that's correct, nesting is disabled when `nesting-level` is 0, and only top-level comments will be displayed. In this case, the "reply" links on top-level comments should probably not be shown. I will add this.

pozitron57:
OK, nice! For `nesting-level = 0` maybe it is better that button `reply` will paste the text of the commenter to the comment form  with comment with `>` prefix? [Juvia](http://juvia-demo.phusion.nl/admin/sites/1/test) (login `discard@phusion.nl`, pass `123456`) does it. If you click `reply to this comment`, you 'll have this in your comment form:

    *Nickname wrote*:
    > text of the comment.

But Isso doesn't show preview on-the-fly and for anyone who isn't familiar with markdown syntax it might be a bit confusing.

sprin:
> OK, nice! For nesting-level = 0 maybe it is better that button reply will paste the text of the commenter to the comment form with comment with > prefix?

I think this is fairly common behavior with single-threaded discussion/comment systems, but I think it should be implemented in a another PR as separate option, eg. "quoted-replies".

I think this PR is ready to be considered for merge. The failing check is related to the installed misaka version, which is resolved by either #212 or #214.

ix5:
See https://github.com/posativ/isso/issues/210#issuecomment-1001245602 and https://github.com/posativ/isso/issues/79

Maybe opinions on this have changed, but allowing more nesting levels would also require a server-side change.

ix5:
Nesting will continue to be limited to one level deep.

PR #228, opened 2016-01-25
config.js error.

smileboywtu:
config.js with wrong path for text.js.

you also miss the make check step in the install from source tutorial, the user should hold the pyflakes
package.

finally the package download from the pipy not work, maybe the wrong version with some dependent package: misaka==1.0.2


posativ:
Cherry-picked here: https://github.com/posativ/isso/commit/b48ac000b170373a98c27b04854838ed209b130c

PR #236, opened 2016-02-25
Create man page directory before moving files to them.

jelmer:
WIthout this, building the manpages fails for me as git doesn't version empty directories.


posativ:
Thanks!

Issue #27, opened 2013-11-05
override thread discovery with custom data-attributes


posativ:

dashohoxha:
This is a big problem for me, because I want to use comments in a mobile app, where url almost never changes, but the content changes (through ajax). So, I need to be able to reference the comments by id, not by url.

Without knowing the structure of the DB and without looking at the code, I would suggest to just get rid of URLs everywhere. We don't need them, we need just IDs in order to identify comment threads. By ID I mean a unique string, not a number. In special cases it can be set to be the url itself, in special cases it can be a number (converted to string), but let the website decide what the ID of the comment thread should be.

postblue:
It was working with 0.9.10, but since 0.10.1 it seems not, too bad!

posativ:
What's the output of `pip --version`?

posativ:
This issue is still open, what do you mean exactly?

blatinier:
This should be closed since you merged #252 right?

rawsh:
Can this be used now? AMP pages are a usecase

ix5:
Should be fixed by #252

PR #252, opened 2016-06-01
Fix issue #27 on override thread discovery with custom data-attributes

blatinier:
Hello,

Here is a fix for the use of data-isso-id.
The problem was that thread ids needs to be URL because of an http curl used to check title of the page. 
I changed this behaviour a little by saying that if you provide a title (using the data-title documented in the client configuration), we don't need to send a curl request to get the title and just grab it. Incidently it makes custom thread id work since the curl to get the title is the only reason thread id is an URL (at least I didn't was any other need for this constraint).

Tell me if any changes needs to be made to the PR.

### DIFF
--- a/docs/docs/configuration/client.rst
+++ b/docs/docs/configuration/client.rst
@@ -107,11 +107,10 @@ Enable or disable voting feature on the client side.
 data-isso-id
 ------------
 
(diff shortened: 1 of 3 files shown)


posativ:
Cool, thanks! Manually adding the title is really something that has been missing. I'll test the custom thread id later.

Issue #10, opened 2013-09-09
admin interface


posativ:

mkzero:
*bump* Still no progress?

danikvsn:
Have you given thought about using Django for your project? It automatically sets up an administrator interface that you can customize

daniyyz:
Hi All, i was looking into the notification.py file, where the url's are generated for Delete or Activate emails, for moderated comments. 

It seems, for this topic, the simplest admin interface would be a looping through of all those URL's and turning them into HTML. then we can click and do stuff via a webpage. 

I'm new to the code, so don't know yet the best place for something like that, but i'll keep poking around.

Thx,
m

PR #256, opened 2016-07-01
Add a basic admin interface (Fix issue #10)

blatinier:
Here is a basic admin interface implementation.
Any comments is welcome.

Access admin through http://my-isso.installation.com/admin
Set password in section `[general]` with the key `admin_password = mylongpasswordwontgetbroken`
Jinja2 is a new dependency for templating admin interface.


pozitron57:
Thank you for this! I've tested it, works fine and interface is OK.

Some things I'd like to see in the future:

- Possibility to group comments by thread. At least indicate thread for each comment.
- Not only `Delete` but also `Edit` button

I am really glad that Isso now has a decent administration panel! Now the only feature I miss in Isso is an email notification of the commenters.

sebix:
@pozitron57 Better add your wishlist to #10

blatinier:
I rebased on the last master and added some of the suggestions of @pozitron57.

blatinier:
List of suggestions by @pozitron57 and their status on this PR:

- [ ] Fix error 500: At right of the panel I see Pages: 0 ». Clicking the link » gives a page of Internal Server Error
- [ ] Possibility to group comments by thread
- [ ] Indicate thread for each comment
- [ ] Not only Delete but also Edit button
- [ ] Possibility to revert changes
- [ ] Show email of the commenter
- [ ] Make everything editable — email, time, name, website, etc.
- [ ] Possibility to reply to the comment
- [ ] Show avatars

blatinier:
@pozitron57 what did you meant by "Possibility to revert changes"?

pozitron57:
@blatinier, imagine that user have deleted comment. And then realized that actually he didn't want to delete it. With the current implementation it is not possible to revert this action — comment is gone. In gmail, for example, when you delete the message, you will have a tooltip: «The conversation has been moved to the Bin.  Undo». Clicking on `Undo` will restore the message. It would be great one will have this `Undo` button after deleting/editing the comment in the admin panel.

By the way, I have one more suggestion: for each comment add a link to the page where the comment is posted.

blatinier:
Editable email, author, website and message.
Things you cannot edit: Thread, time and votes.

blatinier:
I added the link to the thread, which will allow to go answer if the administrator wants to. Having done this, I don't see the real benefit of deplucating the reply mechanism in the admin interface. So I won't do it unless someone convince me it makes a real difference.

jtanx:
Nice work! Forgive me if I'm missing something, but I had to make the following modifications to get it to work:

https://github.com/blatinier/isso/compare/issue-10/admin-interface...jtanx:issue-10/admin-interface?expand=1

blatinier:
Nop you're right. If make a PR on my fork I'd be glad to merge

blatinier:
@posativ Any suggestion on what's mandatory for merging this?
All proposed features are not implemented but most important ones are and it could benefit a lot to users having a real administrative space.

blatinier:
Just rebased on master, to allow some people interested in the admin interface to be up to date regarding master.

hamoid:
If I installed using `pip install isso`, how could I try your changes?

blatinier:
@hamoid sorry for the late answer.
You should install another version of isso using my repository. You clone it, go inside the folder and run `pip install .`
However using virtualenv is not an option if you want to try my version without messing with the current one you have running. And you should be careful to the used ports, so that they don't overlap.

If it seems like too much trouble, try to get @posativ to merge this, but he seems quite busy and do not spend so much time around here from what I saw lately. And given the size of the PR I get that it should be reviewed thouroughly.

Issue #124, opened 2014-08-14
Is it possible to set up isso so Name is mandatory instead of optional?


didibus:
Hi,

I was wondering if there was a configuration option to make it so the Name field is mandatory instead of optional?

Thank You.

tasmo:
+1

rolisz:
This would be nice feature.

Konzertheld:
This has been implemented and can be closed. :)

PR #257, opened 2016-07-05
Add an option for mandatory author (Fix issue #124)

blatinier:
I mirrored the way require-email worked.


posativ:
Looks good to me, thanks!

PR #260, opened 2016-07-16
Vote levels

panta82:
https://github.com/posativ/isso/commit/b0f8606b0f1484f4bea6f0e0f9235247cb0d0766 removes hard-coded behavior that vote count must be hidden when the value is 0. This is now done through CSS.

Use case:
- Ability to disable the default behavior of hiding 0 votes (using CSS overrides)

The other two commits introduce the ability to attach custom classes on comments based on the number of votes. See the changes in documentation for more details.

Use cases:
- Red number when votes < 0, green when >0
- Collapse or gray out comments with many negative votes
- Award good comments with visual flair (eg. a different background color)

### DIFF
--- a/docs/docs/configuration/client.rst
+++ b/docs/docs/configuration/client.rst
@@ -18,6 +18,7 @@ preferably in the script tag which embeds the JS:
             data-isso-avatar-bg="#f0f0f0"
             data-isso-avatar-fg="#9abf88 #5698c4 #e279a3 #9163b6 ..."
             data-isso-vote="true"
+            data-vote-levels=""
             src="/prefix/js/embed.js"></script>
 
 Furthermore you can override the automatic title detection inside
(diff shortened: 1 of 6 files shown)


pozitron57:
For some reason I wasn't able to make this work. I edited manually my `embed.dev.js` to match the proposed changes. Comments without votes do have the class `isso-no-votes`, but comments with votes do not get the class `isso-vote-level-*`. What I may be doing wrong?  Here is my [embed.dev.js](https://comments.lisakov.com/js/embed.dev.js).

panta82:
@pozitron57 

Did you configure vote levels in the script tag? 
As per: https://github.com/panta82/isso/blob/fd3f6f820d15e09153b89d9385a7443845425b6b/docs/docs/configuration/client.rst

Eg:

```
<script data-isso="/prefix/"
        data-vote-levels="-5,5,15"
        src="/prefix/js/embed.js"></script>
```

pozitron57:
Thanks for the reply, I was missing this configuration. I added it but it didn't help. Probably I messed something up while editing embed.dev.js, I'll look into it.

PR #261, opened 2016-07-17
Users

panta82:
Add a limited support for user accounts. A way for the site owner to:
- stylize their own comments
- do the same for moderators/contributors/friends
- prevent anyone else from posting in their name.

For majority of posters, everything works the same as before.

See docs and configuration comments for more details.

TODO-s and remarks:

- Password is transmitted with each POST. Not a big deal for now, but users should probably be warned they should switch to HTTPS and/or not use their shared password.

- I added some CSS and js for better validation + bugfixes.

- Translated the word Password using google translate. If there are translators around, they should give it a look.

- Server doesn't know how to return a proper JSON formatted error. Look into that.

- Added the most basic server-side validation for email. This is needed so that anon users can't impersonate a named user.


posativ:
This is a pretty neat solution, I'll need to review it more thoroughly. I don't like the user management in Isso, it should be solved by $framework a level above (which may provide pluggable user management, whatever. Not a realm of Isso IMHO).

panta82:
@posativ 

Yeah, the real deal would be adding an extensible user support, where the current anon system is just the "default provider" or something. I can also see how any user system would detract from the KISS style of this package.

However, the inability to stylize and protect my own comments (which is expected behavior in any commenting system) was a deal breaker for me. So I added this simple thing for my own purposes.

StructByLightning:
Any update on this? Styling mod comments and preventing impersonation is a really useful feature.

isso/js/app/dom.js
```diff
@@ -32,8 +32,16 @@ define(function() {
         /**
          * Shortcut for `Element.addEventListener`, prevents default event
          * by default, set :param prevents: to `false` to change that behavior.
+         * You can also provide an array of types, to listen on multiple events
          */
         this.on = function(type, listener, prevent) {
+            if (Array.isArray(type)) {
+                var that = this;
+                type.forEach(function (type) {
```
brad: Shouldn't a different var be used for the internally-scoped `type`? Even if just for clarity...

brad:
This is really nice, love the additional error highlighting, that it something I thought was missing. And the user system is so simple, I don't think there is a need for a full fledged framework user system. These are must have changes!

PR #281, opened 2016-10-07
Adding Persian translation.

aidin36:


posativ:
I've used the plural forms for most languages (n != 1), the [Localization Guide](http://docs.translatehouse.org/projects/localization-guide/en/latest/l10n/pluralforms.html)  says something different for Persian. If that matters, please leave a comment :)

PR #289, opened 2016-11-29
Add files for Cdnjs

SnorlaxYum:
@posativ Why not add the js files and make them hosted by CloudFlare? As we know, the js files r almost the same for everyone, so it could be hosted by cdnjs to make it faster to load globally. Then we only have to wait for the isso data from our VPS to load.
https://github.com/cdnjs/cdnjs/pull/9777


SnorlaxYum:
The files embed.js and count.js r the original files which I think r not minified enough.  
The files embed.min.js and count.min.js r minified based on the original files.

posativ:
How do you do versioning when the server version requires a different/new script file? How do I update the JS files when there is a new version? How do I rollback updates, if I messed up?

Is there really a use case to use a CDN to include a random version which may or may not work with the server version? If you think that the static file might be a problem for the VPS performance, a dedicated database server is an actual improvement (which is also not implemented). If the host is too slow to serve some kilobytes of text, you can still use your favourite web server with improved file serving.

SnorlaxYum:
Just fork and send a pull request if u like to. As far as I know, it's possible for several versions to co-exist on that server. Choose wisely for your installed version. I think rollback is possible with just some clicks on the delete button.  
It would be nice to host the js on the CDN instead of a VPS, yep u still need a VPS to run ur isso. However u only need to wait for the data on ur own server. The js files r almost the same for everyone (Different versions can be solved by creating another versions).   
When it's possible to host that on a cdnjs, why not? Actually it will reduce ur server load a little and improve the speed of the js globally much!

PR #291, opened 2016-12-17
fix: add missing i18n entry

ivegotasthma:
Fixes an indexing error a user gets when he tries to make the language of isso to `bg`


ivegotasthma:
@posativ ping, because this is a fatal error for 50% of my site.

PR #293, opened 2016-12-29
added FreeBSD rc.d script

ckoepp:
created a rc.d script for FreeBSD startup of isso within a jail or upon a regular host


posativ:
Cool, thanks!

PR #299, opened 2017-01-07
Danish translation

MadsRC:
Not much to say. I've added a da.js file that should hopefully do :)


jelmer:
You'll also need to add 'da' in various places in isso/js/app/i18n.js

PR #302, opened 2017-01-31
Verify the from email address is set to a valid value

MorrisJobke:
* checks if the `from` value is properly set
* tries to set a name if only the email address is given and checks there if the email address at least contains an `@` and no `space`
* includes some tests

### DIFF
--- a/isso/config.py
+++ b/isso/config.py
@@ -149,8 +149,21 @@ def load(default, user=None):
             logger.info("Your `session-key` has been stored in the "
                         "database itself, this option is now unused")
 
-    if not parseaddr(parser.get("smtp", "from"))[0]:
-        parser.set("smtp", "from",
-                   formataddr(("Ich schrei sonst!", parser.get("smtp", "from"))))
+    from_address = parser.get("smtp", "from").strip()
+
+    if from_address != '' and parseaddr(from_address) == (None, None):
+        if '@' not in from_address:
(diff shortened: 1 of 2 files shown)


MorrisJobke:
@posativ Is there no travis running here? I ran tox locally, but not for all versions.

@noqqe Maybe this is also the reason why your email config doesn't work. Just check if you either use `abc <abc@example.org>` or `abc@example.org`. 😉

skorokithakis:
I'm afraid this pull request is misquided, as a space is a valid character in an email address.

MorrisJobke:
> I'm afraid this pull request is misquided, as a space is a valid character in an email address.

Really? Is it then encoded like `%20`?

skorokithakis:
No, it's enclosed in quotes, like this: valid."email address@"@example.com.

Even this is valid:

"very.(),:;<>[]\".VERY.\"very@\\ \"very\".unusual"@strange.example.com

More info: https://en.wikipedia.org/wiki/Email_address

jelmer:
Could you perhaps simplify this to just print a warning if fromaddr is empty but parseaddr() on it fails? That should be enough of a hint to the user and is still better than silently falling back to "Ich schrei sonst!"

jelmer:
Closing this out for now, since it's stalled and not ready to be merged. Please reopen if there are updates.

PR #303, opened 2017-02-02
Highlight links by underline in comments

MorrisJobke:
Before:

bildschirmfoto 2017-02-02 um 11 35 12

After:

bildschirmfoto 2017-02-02 um 11 34 51

cc @posativ

### DIFF
--- a/isso/css/isso.css
+++ b/isso/css/isso.css
@@ -3,7 +3,7 @@
     -moz-box-sizing: border-box;
     box-sizing: border-box;
 }
-#isso-thread a {
+#isso-thread .isso-comment-header a {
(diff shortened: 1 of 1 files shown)


mdosch:
Actually for me links are properly highlighted according to my blogs css.

auswahl_001

MorrisJobke:
@mdosch this pull request only specifies the rule to only be applied to the header (vor 2 Monaten) and allows an easier theming according the other styles that are loaded on the page. So actually it doesn't add a rule for underline but removes the one that overwrites the default rule for links

posativ:
Looks good to me. Thanks!

PR #311, opened 2017-03-19
Have client read out shared settings from server

pellenilsson:
Some settings (reply-to-self and require-email) always need to be set to the same value on server and client side for correct operation. This change removes the need for such redundant information by having the client read out those settings from the server.

According to the docs this should be true for require-author as well, but that settings doesn't seem to be implemented.

### DIFF
--- a/docs/docs/configuration/client.rst
+++ b/docs/docs/configuration/client.rst
@@ -11,10 +11,6 @@ preferably in the script tag which embeds the JS:
             data-isso-css="true"
             data-isso-css-url="null"
             data-isso-lang="ru"
-            data-isso-reply-to-self="false"
-            data-isso-require-author="false"
-            data-isso-require-email="false"
-            data-isso-reply-notifications="false"
             data-isso-max-comments-top="10"
             data-isso-max-comments-nested="5"
             data-isso-reveal-on-click="5"
@@ -128,26 +124,6 @@ languages. The language is configured by its `ISO 639-1
 You find a list of all supported languages on `GitHub
 <https://github.com/posativ/isso/tree/master/isso/js/app/i18n>`_.
 
-data-isso-reply-to-self
------------------------
-
-Set to `true` when spam guard is configured with `reply-to-self = true`.
-
-data-isso-require-author
-------------------------
-
-Set to `true` when spam guard is configured with `require-author = true`.
-
-data-isso-require-email
------------------------
-
-Set to `true` when spam guard is configured with `require-email = true`.
-
-data-isso-reply-notifications
------------------------------
-
-Set to `true` when reply notifications is configured with `reply-notifications = true`.
-
 data-isso-max-comments-top and data-isso-max-comments-nested
 ------------------------------------------------------------
 
@@ -164,7 +140,9 @@ Number of comments to reveal on clicking the "X Hidden" link.
 data-isso-avatar
 ----------------
 
-Enable or disable avatar generation.
+Enable or disable avatar generation. Ignored if gravatar is enabled on
+server side, since gravatars will take precedence and disable avatar
+generation.
 
 data-isso-avatar-bg
 -------------------
@@ -178,14 +156,6 @@ Set avatar foreground color. Up to 8 colors are possible. The default color
 scheme is based in `this color palette <http://colrd.com/palette/19308/>`_.
 Multiple colors must be separated by space. If you use less than eight colors
 and not a multiple of 2, the color distribution is not even.
-
-data-isso-gravatar
-------------------
-
-Uses gravatar images instead of generating svg images. You have to set
-"data-isso-avatar" to **false** when you want to use this. Otherwise
-both the gravatar and avatar svg image will show up. Please also set
-option "gravatar" to **true** in the server configuration...
 
 data-isso-vote
 --------------
@@ -212,3 +182,38 @@ data-isso-feed
 Enable or disable the addition of a link to the feed for the comment
 thread. The link will only be valid if the appropriate setting, in
 ``[rss]`` section, is also enabled server-side.
+
+
+
+Deprecated Client Settings
+==========================
+
+In earlier versions the following settings had to mirror the
+corresponding settings in the server configuration, but they are now
+read out from the server automatically.
+
+data-isso-reply-to-self
+-----------------------
+
+Set to `true` when spam guard is configured with `reply-to-self = true`.
+
+data-isso-require-author
+------------------------
+
+Set to `true` when spam guard is configured with `require-author = true`.
+
+data-isso-require-email
+-----------------------
+
+Set to `true` when spam guard is configured with `require-email = true`.
+
+data-isso-reply-notifications
+-----------------------------
+
+Set to `true` when reply notifications is configured with `reply-notifications = true`.
+
+data-isso-gravatar
+------------------
+
+Set to `true` when gravatars are enabled with `gravatar = true` in the
+server configuration.
--- a/isso/js/embed.js
+++ b/isso/js/embed.js
@@ -41,8 +41,10 @@ require(["app/lib/ready", "app/config", "app/i18n", "app/api", "app/isso", "app/
             feedLinkWrapper.appendChild(feedLink);
             isso_thread.append(feedLinkWrapper);
         }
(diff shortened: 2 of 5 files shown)


brad:
This is brilliant! `require-author` is working here, maybe that should be part of the PR as well. I think also documentation changes should be made before this can be merged

jelmer:
Any chance you could resolve the conflicts?

pellenilsson:
In my fork, I took this a step further and centralized **all** settings to the server side (see ed5692181ed856053e7f5420551b3f9c59a883fc).

The main content of my fork is authentication with Facebook, Google and OpenID as described in [this blog post](https://pantarei.xyz/posts/isso-with-social-network-integration/) which also has a live demo.

I was earlier under the impression that this line of work was not desired in mainline Isso, but since this may have changed we could have a discussion about what functionality to include before looking at the conflicts.

pellenilsson:
I meant the entire fork. I just want to avoid multiple merges by deciding what to include.

jelmer:
I think either way we'd want to merge those changes one by one rather than all at once.

pellenilsson:
Yes, moving settings is a quite separate change of course. But i thought we could agree on moving all settings or just the "redundant" settings that now have to be specified on both server and client side.

On the other hand, moving only the "redundant" settings is easier because there are no backward compatibility issues. Moving all settings would either require supporting the old way as well, or forcing users to move their settings.

blatinier:
Bumping isso version to 2.x.x might be enough to indicate breaking change don't you think? 
The centralization of config is really neat and would be very much appreciated since it would simplify the process of installing/setting up isso.

blatinier:
I fixed conflicts, and it should be ok. Needs a bit of testing though.

TheKerbycz:
Hi, I was going to use the pellenilsson's fork with social auth, but after reading this conversation I'm a not really sure what version (and branch) should I use. Should I use branch 'feature/get-server-config' or 'master'?  Am I right, that the social login feature is only in this fork? If so, are there any plans to include it also in the posativ's repository? Thanks

pellenilsson:
Hi @TheKerbycz! The social auth stuff is only available in the master branch of my forked repo https://github.com/pellenilsson/isso. Since it was forked some time ago it misses some of the newer features. Whether it will be merged to posativ's repo is an open question.

jelmer:
This seems reasonable to me; As it has no unit tests, has anybody else managed to successfully test it?

ix5:
@pellenilsson would you mind rebasing this on top of current master?

While you're at it, there are a few more settings that are shared between client and server and need to be kept in sync. I've quickly glanced over the docs and found these:

- `data-isso-require-author` with `[guard] require-author`
- `data-isso-require-email` with `[guard] require-email`
- `data-isso-reply-notifications` with `[general] reply-notifications`
- `data-isso-gravatar` with `[general] gravatar`

Also, it'd be nice for the "gravatar" setting to trump the regular avatars if set to true instead of having both generated.

@pellenilsson if you're done with the shared settings, we'd need to document this new behaviour as well, first in CHANGES and then also in the docs. I'd be happy to help you there.

---

Just additionally, this is  wildly out of scope, but we might want to display the available HTML tags and markdown options to clients in some kind of way. So exposing those would fall under the same mechanism.

pellenilsson:
@ix5 I merged in latest master and added the additional settings that you mentioned, and tried to handle the gravatar scenario. I also updated the docs.

But I no longer have a setup for testing this out since I haven't worked on Isso for several years, so it would be nice if someone else could do some testing.

isso/js/embed.js
```diff
@@ -57,6 +57,14 @@ require(["app/lib/ready", "app/config", "app/i18n", "app/api", "app/isso", "app/
             config["max-comments-top"],
             config["max-comments-nested"]).then(
             function (rv) {
+                for (var setting in rv.config) {
+                    config[setting] = rv.config[setting]
+                }
+
+                $("#isso-thread").append($.new('h4'));
```
ix5, line 64: These snuck in by accident

docs/docs/configuration/client.rst
```diff
@@ -164,7 +140,8 @@ Number of comments to reveal on clicking the "X Hidden" link.
 data-isso-avatar
 ----------------
 
-Enable or disable avatar generation.
+Enable or disable avatar generation. Ignored if gravatar is enabled on
```
ix5, line 143: Avatars will be disabled in addition to this client setting being ignored.
Please also move the gravatar setting to deprecated settings further down.

ix5 (commented):
Left a few comments. Thanks for rebasing and addressing those issues. I'll try to test thoroughly once I find the time.

ix5:
Pinging @stefangehn @Guts @zackw @vsajip @Konzertheld for testing/review. Would be nice of one of you to take a look at this!

isso/js/embed.js
ix5: Also, would be nice if there was at least a message in the browser console about a client setting being overridden by a server setting. That'd require separating against the client defaults though.

vsajip: When you say "separating against", do you mean keeping both sets of values so as to show the differences in the console message?

ix5: No, I mean we'd need a way to check if the user has attempted to modify a client setting via `data-isso-*` before doing this type of check:

```js
if (config[setting] != rv.config[setting]) {
    console.log("Isso: Client value xyz overridden by server value.");
    console.log("Since Isso version 0.12.6, data-isso-%value is only configured via the server to keep client and server in sync");
    config[setting] = rv.config[setting];
}
```

There might be a discrepancy just because the client is still using default values but the value has been set on the server.

pellenilsson: Ah... In my original commit I moved those lines into the `api.fetch` callback, because the `Postbox` constructor needed the final configuration and therefore had to be called after receiving/updating the config from the server response. The auto-merger screwed up and just added the code instead of moving it, since the code looks a bit different in master now. It still needs to be moved.

pellenilsson: Will fix later this week.

ix5: Please squash while you're at it. Thanks!

docs/docs/configuration/client.rst
pellenilsson: Fixed.

isso/js/embed.js
pellenilsson: Fixed the auto-merger mistake and added the warning message based on the suggestion above (but note the extra condition).

I will wait with squashing until someone has actually run the code, in case more fixes are needed.

ix5:
@pellenilsson currently, we're in a bit of a catch-22: `fetchComments()` checks if `#isso-root` is present, but fails since the call to create a new `h4` header, a `Postbox` and the `div#isso-root` element is part of `fetchComments()` itself.

The fix for that is to add the header and root element early, but only later insert the postbox. I'm not aware of any issues concerning the postbox being inserted _after_ the root element.

I've pushed to your branch with a few suggestions and improvements. If you like, you can reset your HEAD to https://github.com/ix5/isso/tree/pull-311-rebased , I've squashed  the commits for you already there and written a succinct commit message.

pellenilsson:
> I've pushed to your branch with a few suggestions and improvements. If you like, you can reset your HEAD to https://github.com/ix5/isso/tree/pull-311-rebased , I've squashed the commits for you already there and written a succinct commit message.

Thanks! I force-pushed your squashed branch now.

ix5:
@pellenilsson I haven't forgotten your PR, I am just trying to think of ways to verify this does not break anything.
Will try to get to it during the remainder of this week.

ix5:
Tested various configuration disparities between client and server and the server always takes precedence.
I couldn't spot any obvious bugs or oversights.

The warning logic is off for when default client settings (from `config.js`) are overridden by server settings/defaults. But that's for another day to fix, this PR has languished in limbo way too long.

Thank you so much @pellenilsson for this great improvement!

ix5:
@pellenilsson in case you still feel like contributing after such a long wait, have a look at https://github.com/ix5/isso/commits/pull-311-preserve-defaultconf for slightly correcting the warnings in the console.

PR #314, opened 2017-03-23
Use bleach to sanitize HTML; allows use with newer versions of html5lib.

### LINKED ISSUE #296: fyi/ftr: html5lib dependency
Just for your information/for the record:

I upgraded on my server (Arch Linux) to python-html5lib version 0.999999999-1, and since the requirement was not met in `setup.py`, isso stopped working. I then tried to compile isso with the new version of html5lib, but this led to another error:

```
[..]
ImportError: No module named 'html5lib.sanitizer'
```

The solution was to downgrade to python-html5lib version 0.9999999-2 and ignore this package in `/etc/pacman.conf`.

jelmer:
Fixes #296 

This patch is needed to fix the Debian packaging as Debian already ships a newer version of html5lib, which no longer supports sanitizing. I'm submitting this PR in case you also want to ship this upstream.

### DIFF
--- a/isso/tests/test_html.py
+++ b/isso/tests/test_html.py
@@ -63,7 +63,6 @@ class TestHTML(unittest.TestCase):
             print("Hello, World")
             </code></pre>""")
 
-    @unittest.skipIf(html.HTML5LIB_VERSION <= html.HTML5LIB_SIMPLETREE, "backport")
     def test_sanitizer(self):
         sanitizer = html.Sanitizer(elements=[], attributes=[])
         examples = [
@@ -74,9 +73,8 @@ class TestHTML(unittest.TestCase):
             ('<script>alert("Onoe")</script>', 'alert("Onoe")')]
 
         for (input, expected) in examples:
-            self.assertEqual(html.sanitize(sanitizer, input), expected)
+            self.assertEqual(sanitizer.sanitize(input), expected)
 
-    @unittest.skipIf(html.HTML5LIB_VERSION <= html.HTML5LIB_SIMPLETREE, "backport")
     def test_sanitizer_extensions(self):
         sanitizer = html.Sanitizer(elements=["img"], attributes=["src"])
         examples = [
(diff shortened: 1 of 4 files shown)


jelmer:
Tests with Python2.6 fail with this PR, because bleach no longer supports Python 2.6. https://github.com/mozilla/bleach/issues/275

gloomy-ghost:
Since support for py 2.6 has been dropped, can this PR be reconsidered?

jelmer:
I've reopened the pull request here: https://github.com/posativ/isso/pull/483

However, master seems to have a lot of test failures at the moment so those will need to be addressed first.

gloomy-ghost:
@jelmer wow I just noticed you are isso's maintainer for debian as well. The isso package on debian is 1-year-old now, so I've been trying to create an updated package using current master. However, I only realised I will have to do a NMU (plus finding a mentor) after I got most of it working. What do you think would be the most suitable next step for me?

jelmer:
On Fri, Sep 28, 2018 at 08:18:18AM +0000, gloomy-ghost wrote:
> @jelmer wow I just noticed you are isso's maintainer for debian as well. The isso package on debian is 1-year-old now, so I've been trying to create an updated package using current master. However, I only realised I will have to do a NMU (plus finding a mentor) after I got most of it working. What do you think would be the most suitable next step for me?
If you get the package in a reasonable shape for an upload, I'd be
happy to sponsor.

-- 
Jelmer Vernooĳ <jelmer@jelmer.uk>
PGP Key: https://www.jelmer.uk/D729A457.asc

gloomy-ghost:
uploaded at https://mentors.debian.net/package/isso

I couldn't get `r.js` to be happy with debian's jade packed in commonjs, but I suppose it's better than a _use upstream_...

gloomy-ghost:
> I couldn't get `r.js` to be happy with debian's jade packed in commonjs, but I suppose it's better than a _use upstream_...

it works now :smile:

PR #316, opened 2017-03-28
Better configure options for Chinese language

shengbinmeng:
Allow the user configure language to "zh", "zh_CN", "zh_TW", and make them work as expected.

### DIFF
--- a/isso/js/app/i18n.js
+++ b/isso/js/app/i18n.js
@@ -2,8 +2,8 @@ define(["app/config", "app/i18n/bg", "app/i18n/cs", "app/i18n/de",
         "app/i18n/en", "app/i18n/fa", "app/i18n/fi", "app/i18n/fr",
         "app/i18n/hr", "app/i18n/ru", "app/i18n/it", "app/i18n/eo",
         "app/i18n/sv", "app/i18n/nl", "app/i18n/el_GR", "app/i18n/es",
-        "app/i18n/vi", "app/i18n/zh_CN"],
-        function(config, bg, cs, de, en, fa, fi, fr, hr, ru, it, eo, sv, nl, el, es, vi, zh) {
+        "app/i18n/vi", "app/i18n/zh_CN", "app/i18n/zh_TW"],
+        function(config, bg, cs, de, en, fa, fi, fr, hr, ru, it, eo, sv, nl, el, es, vi, zh_CN, zh_TW) {
 
     "use strict";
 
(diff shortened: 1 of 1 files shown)


isso/js/app/i18n.js
```diff
@@ -2,8 +2,8 @@ define(["app/config", "app/i18n/bg", "app/i18n/cs", "app/i18n/de",
         "app/i18n/en", "app/i18n/fa", "app/i18n/fi", "app/i18n/fr",
         "app/i18n/hr", "app/i18n/ru", "app/i18n/it", "app/i18n/eo",
         "app/i18n/sv", "app/i18n/nl", "app/i18n/el_GR", "app/i18n/es",
-        "app/i18n/vi", "app/i18n/zh_CN"],
-        function(config, bg, cs, de, en, fa, fi, fr, hr, ru, it, eo, sv, nl, el, es, vi, zh) {
+        "app/i18n/vi", "app/i18n/zh_CN", "app/i18n/zh_CN", "app/i18n/zh_TW"],
```
jelmer: This seems to add a duplicate entry for zh_CN, did you mean to add just "app/i18/zh" ?

shengbinmeng: Actually I'd like to make configuration items "zh"," zh_CN", "zh_TW" correspond to the files "app/i18n/zh_CN", "app/i18n/zh_CN", "app/i18n/zh_TW", respectively (if I understood it right). In other words, treat the "zh" as an alias of "zh_CN".

shengbinmeng: Further explanation: the users can configure `lang="zh_CN"` to use the CN version or `lang="zh_TW"` to use the TW version; and they can also simply configure `lang="zh"` to use the CN version (Simplified Chinese).

jelmer: that makes sense to me, but wouldn't you want "app/i18n/zh_CN" *and* "app/i18n/zh" in the list above rather than "app/i18n/zh_CN" twice?

shengbinmeng: Hi, `app/i18n/zh_CN` is the translation file (located here: https://github.com/posativ/isso/tree/master/isso/js/app/i18n), not an entry. That piece of code means entry "zh" and " zh_CN" both use the file `app/i18n/zh_CN` (which is on purpose). I think the diff format makes it a little confusing, please see the whole JS function or file to get a better view.

Issue #82, opened 2014-04-22
Identifying the author of the document


programLyrique:
It would be useful to identify the comments of author of the document (or, if it is difficult to do it per document, per website) which is commented, by looking at the email and the username for instance, with a chosen color for the background of the comments for example (and maybe some indication next to the username and some enhancement for the avatar).
![isso](https://cloud.githubusercontent.com/assets/382128/2765131/5957c776-ca20-11e3-9700-bc2aea21e436.png)
.

ix5:
Closing in favour of discussion in #321.

Especially https://github.com/posativ/isso/pull/321#issuecomment-382148195 is a nice solution.

ix5:
Re-opening until someone adds the proposed CSS solution to the docs.

BBaoVanC:
I think a badge next to the author's name plus a different name color would be better than a message background. I think it's easier than trying to deal with trying to make the background fit in with the visual style.

ix5:
As a first step to allow better styling, the author `hash` should be inserted into the `.isso-comment` div, not only as the has to the svg.

@BBaoVanC would you like to work on something like this?

BBaoVanC:
> As a first step to allow better styling, the author `hash` should be inserted into the `.isso-comment` div, not only as the hash to the svg.
> 
> @BBaoVanC would you like to work on something like this?

I'll try and see what I can do. I haven't ever done much with JS so for that part of the client I might need help. I'll make a draft PR if I can figure something out.

ix5:
That's what I mean:

```diff
diff --git a/isso/js/app/templates/comment.js b/isso/js/app/templates/comment.js
index d0daf99..53738b2 100644
--- a/isso/js/app/templates/comment.js
+++ b/isso/js/app/templates/comment.js
@@ -9,7 +9,7 @@ var html = function (globals) {
   var author = comment.author ? comment.author : i18n('comment-anonymous');
 
   return "" +
-"<div class='isso-comment' id='isso-" + comment.id + "'>"
+"<div class='isso-comment' id='isso-" + comment.id + "' data-hash='" + comment.hash + "'>"
 + (conf.gravatar ? "<div class='isso-avatar'><img src='" + comment.gravatar_image + "'></div>" : '')
 + (conf.avatar ? "<div class='isso-avatar'><svg data-hash='" + comment.hash + "'</svg></div>" : '')
 + "<div class='isso-text-wrapper'>"
```

[...]

PR #321, opened 2017-04-20
Highlighting for author comments

Libbum:
I've completed #82 and made it into a feature rather than a manual hack.

Under the [general] section in isso.conf there is now an `author` variable that users can set to an email which they use on their websites to differentiate themselves from other commenters.

issoauthor

There were a few additions needed to get this done. 
1. Hash the author address on startup.
2. Extend the api to allow `/author` requests (which returns the hash).
3. Append the api response to the `conf` struct on the client side.
4. Compare the comment and blog author hashes, append a new css selector `isso-highlight`.

I guess this implementation has no issue with using a users' IP too, although I haven't tested that. In fact, I haven't written any test in yet. Thought I'd get some feedback to see if this was something of interest first.

### DIFF
--- a/isso/__init__.py
+++ b/isso/__init__.py
@@ -90,6 +90,7 @@ class Isso(object):
         self.signer = URLSafeTimedSerializer(self.db.preferences.get("session-key"))
         self.markup = html.Markup(conf.section('markup'))
         self.hasher = hash.new(conf.section("hash"))
+        self.author = self.hasher.uhash(conf.get('general', 'author'));
 
         super(Isso, self).__init__(conf)
 
--- a/isso/css/isso.css
+++ b/isso/css/isso.css
@@ -23,6 +23,11 @@
     color: #AAA;
 }
 
+.isso-highlight {
+    padding: 0em 0em 0.5em 0.5em;
+    background-color: rgba(200, 200, 200, 0.4);
+}
+
 .isso-comment {
     max-width: 68em;
     padding-top: 0.95em;
--- a/isso/js/app/api.js
+++ b/isso/js/app/api.js
@@ -177,6 +177,20 @@ define(["app/lib/promise", "app/globals"], function(Q, globals) {
         return deferred.promise;
     };
 
+    var author = function() {
+        var deferred = Q.defer();
+        curl("GET", endpoint + "/author", null, function(rv) {
+                if (rv.status === 200) {
+                    deferred.resolve(JSON.parse(rv.body));
+                } else if (rv.status === 404) {
+                    deferred.resolve({author: ""});
+                } else {
+                    deferred.reject(rv.body);
+                }
+            });
+        return deferred.promise;
+    };
+
     var like = function(id) {
         var deferred = Q.defer();
         curl("POST", endpoint + "/id/" + id + "/like", null,
@@ -201,6 +215,7 @@ define(["app/lib/promise", "app/globals"], function(Q, globals) {
         view: view,
         fetch: fetch,
         count: count,
+        author: author,
         like: like,
         dislike: dislike
     };
--- a/isso/js/app/isso.js
+++ b/isso/js/app/isso.js
@@ -5,6 +5,8 @@ define(["app/dom", "app/utils", "app/config", "app/api", "app/jade", "app/i18n",
 
     "use strict";
 
+    api.author().then(function(rv) { config["blogauthor"] = rv; });
+
     var Postbox = function(parent) {
 
         var localStorage = utils.localStorageImpl,
--- a/isso/js/app/text/comment.jade
+++ b/isso/js/app/text/comment.jade
@@ -1,4 +1,4 @@
-div(class='isso-comment' id='isso-#{comment.id}')
+div(class=(comment.hash == conf.blogauthor) ? 'isso-comment isso-highlight' : 'isso-comment' id='isso-#{comment.id}')
     if conf.avatar
         div(class='avatar')
             svg(data-hash='#{comment.hash}')
--- a/isso/views/comments.py
+++ b/isso/views/comments.py
@@ -82,6 +82,7 @@ class API(object):
         ('new',     ('POST', '/new')),
         ('count',   ('GET', '/count')),
         ('counts',  ('POST', '/count')),
+        ('author',  ('GET', '/author')),
         ('view',    ('GET', '/id/<int:id>')),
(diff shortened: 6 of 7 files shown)


jelmer:
Thanks for working on this.

Having author as part of the configuration means you can only have one author per isso instance, and adds an extra HTTP request. Wouldn't it make more sense to have a client setting for the author, so that it can change on a per-thread basis?

Libbum:
I certainly agree on the idea, but couldn't quite figure out the best way to obfuscate user info. Here we can hash the authors email address / url and match against that so we don't have to worry about crawlers/spammers. If there's a good way to do this client side I'd be happy to change the implementation and extend the capability such that we can invoke per-thread authors. Do you have any suggestions to that end?

jelmer:
The obfuscation would have to be done by whatever sets data-isso-author; i.e. we'd specify that it can be set to a hashed version of the authors' email address.

posativ:
During development I had a JavaScript function that produces the same results as the current hashes (hashed with PBKDF2) generated from the server. The only downside is like multiple hundred milliseconds to generate this hash.

Libbum:
(sorry for the late reply, I've been on vacation)

> The obfuscation would have to be done by whatever sets data-isso-author; i.e. we'd specify that it can be set to a hashed version of the authors' email address.

This is exactly what's happening already, and it's the server side that does this. 
Can you think of a better process for the user here?

Sticking to just one thread, one author for now, and using the current isso implementation; this is what I see a client side method being:

1. The author knows their email adress, but does not want it posted in plain text in each post.
2. Post a test comment.
3. View source and find their user hash.
4. Delete test comment.

Now, with their obtained hash and an author highlight system enabled:

5. Assign their hash to a css value in the isso class settings.

If you add in multiple authors, now each author needs to know their hash and each thread needs css descriptors (currently, the css in isso global).

As someone who can work of development of a project like this, that's not too bad - and is exactly what I've done on my blog; but I think it's a tall order to ask this of regular users.

One option would be to extend the cli portion of isso to spit out a hash given a set of credentials. This should be run during setup and then can be set manually in css. I don't really see the need to have client side hashing, as that means the implementation would give the client the author credentials to hash and make obfuscation a moot point.

Multiple authors probably requires database extensions too now that I think about it, or a lot of manual intervention from the user side. So I think for the moment, I'd like to solve a single author problem first, with view of extending it to multiple authors later if possible.

So with that in mind, I still agree with your points about minimising server requests, but still can't see a rational way of moving this client side AND keeping author credentials private.

jelmer:
On Fri, Aug 04, 2017 at 09:46:21AM -0700, Tim DuBois wrote:
> (sorry for the late reply, I've been on vacation)
> 
> > The obfuscation would have to be done by whatever sets data-isso-author; i.e. we'd specify that it can be set to a hashed version of the authors' email address.
> 
> This is exactly what's happening already, and it's the server side that does this. 
> Can you think of a better process for the user here?
> 
> Sticking to just one thread, one author for now, and using the current isso implementation; this is what I see a client side method being:
> 
> 1. The author knows their email adress, but does not want it posted in plain text in each post.
> 2. Post a test comment.
> 3. View source and find their user hash.
> 4. Delete test comment.
> 
> Now, with their obtained hash and an author highlight system enabled:
> 
> 5. Assign their hash to a css value in the isso class settings.
> 
> If you add in multiple authors, now each author needs to know their hash and each thread needs css descriptors (currently, the css in isso global).
> 
> As someone who can work of development of a project like this, that's not too bad - and is exactly what I've done on my blog; but I think it's a tall order to ask this of regular users.
> 
> One option would be to extend the cli portion of isso to spit out a hash given a set of credentials. This should be run during setup and then can be set manually in css. I don't really see the need to have client side hashing, as that means the implementation would give the client the author credentials to hash and make obfuscation a moot point.
> 
> Multiple authors probably requires database extensions too now that I think about it, or a lot of manual intervention from the user side. So I think for the moment, I'd like to solve a single author problem first, with view of extending it to multiple authors later if possible.
> 
> So with that in mind, I still agree with your points about minimising server requests, but still can't see a rational way of moving this client side AND keeping author credentials private.  
There is certainly an argument to be made for both approaches. If the hash can
be provided client-side, then that allows for the author to be
different on a per-thread basis. But as you point out, that
makes the setup more complex for simpler use cases that could do with
just a setting in the configuration file.

If you're going to implement this server side then it would
at least be nice to avoid the extra complexity on the client
side and the extra HTTP request. You could just include a "highlight"
field in the JSON representation of comments that could be set by the
server and then be used for determining highlighting by the client code.

I also think the configuration variable shouldn't just say "author" in
that case, but perhaps take a list of email addresses from people to
highlight. That at least doesn't suggest that isso can't be used with
multi-author blogs.

(In the future, it would still be possible to *also* provide a list of hashes
to high-light on the client side)

Cheers,

Jelmer

Libbum:
Those are good ideas. I'll have a think about implementations and update this when I get the time.

vincentbernat:
FI, I am using this bit of CSS:

```css
    #isso-thread .avatar > svg[data-hash="0bb73c4f5196"] {
        box-shadow: 0 0 12px #B80F28;
        border-radius: 50%;
    }
```

I would have preferred to add a kind of badge, but the hash is on the SVG which doesn't have a relation with the header of the comment, so with just CSS, I don't think it's possible add a badge. And I didn't try better because it's good enough for me.

StructByLightning:
@vincentbernat Nice bit of CSS, that was exactly what I was looking for. It's kind of a hack though and pretty limited, so I think having builtin support for owner styling is still useful.

jelmer (changes requested):
per earlier comments

ix5:
For the record, this is what works to generate the hash:
```python
import codecs
from hashlib import pbkdf2_hmac as pbkdf2

val = "admin@example.com".encode("utf-8")

# Default parameters, change if you have different config:
salt = b"Eech7co8Ohloopo9Ol6baimi"
iterations = 1000
dklen = 6
func = "sha1"

compute = pbkdf2(
    hash_name=func,
    password=val,
    salt=salt,
    iterations=iterations,
    dklen=dklen
)

hash_ = compute

uhash = codecs.encode(hash_, "hex_codec").decode("utf-8")

print("uhash: %s" % uhash)
```

---

Closing this since the OP has seemingly abandoned and not reacted to jelmer's review comments. Maybe one day a similar PR can find its way into the project.

Issue #319, opened 2017-04-15
Feature request: Reverse-sort the comments on rendering (latest first)


PeterCxy:
It's not very important, but I do think that sorting comments in 'latest-first' order will be more natural for readers. Maybe an option 'data-isso-sorting' can be added to the JavaScript client.

dbrgn:
I agree. Seems to be a duplicate of #4.

blatinier:
Agree. Closing. @jelmer I add the label feature to #4

PR #322, opened 2017-04-22
Implement customizable sorting

### LINKED ISSUE #265: How to set the order of the comments?
Generally, the newest is on the bottom.  
Instead I want the newest to be on the top, how to do it?

dbrgn:
In the first commit, I implemented sorting by oldest or newest. The implementation is very simple and would fix #265 and #319. I then extended the sorting algorithm in the second commit to support multiple sorting methods and added the "upvote" sorter. With this, #4 is also fixed.

To see how it works, the updated part of the documentation:

> A comma-separated list of thread sorting methods that are applied in the
> specified order.
> 
> Possible sorting methods:
> 
> - `newest`: Bring newest comments to the top
> - `oldest`: Bring oldest comments to the top
> - `upvotes`: Bring most liked comments to the top
> 
> You can combine sorting methods. If you specify `upvotes,newest`, then comments
> will be sorted by upvotes, and comments with the same number of upvotes will be
> sorted by date (newest first).
> 
> Note that only the top level threads are sorted according to this
> configuration. The thread replies are still sorted in chronological order
> (oldest first), so that they can be read from top to bottom.
>
> Default sorting is `oldest`.

I kept the default sorting at `oldest`, but you might want to change it to `upvotes,newest`.

### DIFF
--- a/docs/docs/configuration/client.rst
+++ b/docs/docs/configuration/client.rst
@@ -22,6 +22,7 @@ preferably in the script tag which embeds the JS:
             data-isso-avatar-fg="#9abf88 #5698c4 #e279a3 #9163b6 ..."
             data-isso-vote="true"
             data-isso-vote-levels=""
+            data-isso-sorting="oldest"
             data-isso-feed="false"
             src="/prefix/js/embed.js"></script>
 
@@ -149,6 +150,11 @@ For example, the value `"-5,5"` will cause each `isso-comment` to be given one o
 - `isso-vote-level-2` for scores of `5` and greater
 
 These classes can then be used to customize the appearance of comments (eg. put a star on popular comments)
+
+data-isso-sorting
+-----------------
+
+This can be set either to `oldest` (the default) or `newest`.
 
 data-isso-feed
 --------------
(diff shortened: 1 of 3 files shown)


dkmooers:
Can this PR be merged? I'd really love to have this feature personally and the work is already done!

dkmooers:
Actually, this doesn't work with auto-hiding of overflow comments, since the sorting happens on the client. Let's say I have 500 comments on one post, and I limit the display to 10 top-level comments, and I want to display NEWEST first: the embed.min.js script grabs the 10 oldest comments from the Isso server, then sorts _those_ by newest. So I guess the sorting needs to happen on the Isso server, rather than the client, to work with hiding overflow comments.

dbrgn:
@dkmooers you're right about that issue. I don't use hiding, so that's a non-issue for me so far.

@jelmer @posativ what would be your preferred approach? can this PR be saved somehow, or would everything need to be done on the server side?

jelmer:
Yes, I believe we should push this logic into the server. :-/

dbrgn:
Then I'll close this :slightly_smiling_face:

PR #343, opened 2017-09-04
Support for used-defined log formats

brthanmathwoag:
Hi,

currently only the messages are logged to the log file, which makes it difficult to inspect, as there are no dates or severities to grep for.
I patched the code so that they are included, and also let the user can provide a custom format in the config file.


brthanmathwoag:
Nobody mentioned a thing, but this PR does not make much sense,
as logging in isso is generally broken - only startup messages are written to a log file,
other are only shown in the console, and they are using different formats.

PR #350, opened 2017-10-31
Added hungarian translation

guthypeter:


mpchadwick:
I think this may have broken things?

image

FriedrichFroebel:
Seems like the `hu` parameter is missing at https://github.com/posativ/isso/pull/350/files#diff-038fa9a64e0302ab76c3957100beffbdR6 which causes problems.

Issue #338, opened 2017-07-28
Improve Accessibility of the Comment Textfield


austinseraphin:
The accessibility of the textfield for leaving a comment could use a little improving. A label would help identify it, as with the other fields. Also, for some reason, Firefox on Linux gets confused by it and reads a blank line, but Safari on a Mac has no problem identifying it as a textfield.

ix5:
Some basic approaches mentioned in https://github.com/posativ/isso/pull/353#issuecomment-346217728

ix5:
Should be fixed by https://github.com/posativ/isso/pull/861

PR #353, opened 2017-11-21
Add <label>s to postbox <input>s

mpchadwick:
The labels are [visually hidden](https://css-tricks.com/places-its-tempting-to-use-display-none-but-dont/) since the `placeholder` attribute is clear enough for sighted users. Related to https://github.com/posativ/isso/issues/338


isso/js/app/text/postbox.jade
```diff
@@ -5,13 +5,16 @@ div(class='isso-postbox')
                 = i18n('postbox-text')
         section(class='auth-section')
             p(class='input-wrapper')
-                input(type='text' name='author' placeholder=i18n('postbox-author')
+                label(for='isso-postbox__author',class='isso-visuallyhidden')= i18n('postbox-author')
```
aquatix: Looks good, but can you fix the whitespace around the '=' before i18n? I think adding a space there would be best. *nitpick* :) Same with the two other lines.

mpchadwick:
Upon further research, I think that while this approach will remove a violation during an accessibility scan, I'm not sure it actually the right thing to do for users.

A better approach would be to take the current placeholder text and use it a label **which is not visually hidden** and then either eliminate the use of placeholders entirely, or else use them as the are intended...

> The placeholder attribute represents a short hint (a word or short phrase) intended to aid the user with data entry.

Here are some links about this...

https://developer.paciellogroup.com/blog/2011/02/html5-accessibility-chops-the-placeholder-attribute/
http://www.maxability.co.in/2016/01/placeholder-attribute-and-why-it-is-not-accessible/
https://www.nngroup.com/articles/form-design-placeholders/

PR #355, opened 2017-11-22
Fix hungarian translation

mpchadwick:
#350 introduces an error per https://github.com/posativ/isso/pull/350#issuecomment-345892951

### DIFF
--- a/isso/js/app/i18n.js
+++ b/isso/js/app/i18n.js
@@ -3,7 +3,7 @@ define(["app/config", "app/i18n/bg", "app/i18n/cs", "app/i18n/da",
         "app/i18n/fr", "app/i18n/hr", "app/i18n/ru", "app/i18n/it",
         "app/i18n/eo", "app/i18n/sv", "app/i18n/nl", "app/i18n/el_GR",
         "app/i18n/es", "app/i18n/vi", "app/i18n/zh_CN", "app/i18n/zh_CN", "app/i18n/zh_TW", "app/i18n/hu"],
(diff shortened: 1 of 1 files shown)


isso/js/app/i18n.js
```diff
@@ -3,7 +3,7 @@ define(["app/config", "app/i18n/bg", "app/i18n/cs", "app/i18n/da",
         "app/i18n/fr", "app/i18n/hr", "app/i18n/ru", "app/i18n/it",
         "app/i18n/eo", "app/i18n/sv", "app/i18n/nl", "app/i18n/el_GR",
         "app/i18n/es", "app/i18n/vi", "app/i18n/zh_CN", "app/i18n/zh_CN", "app/i18n/zh_TW", "app/i18n/hu"],
-        function(config, bg, cs, da, de, en, fa, fi, fr, hr, ru, it, eo, sv, nl, el, es, vi, zh, zh_CN, zh_TW) {
+        function(config, bg, cs, da, de, en, fa, fi, fr, hr, hu, ru, it, eo, sv, nl, el, es, vi, zh, zh_CN, zh_TW) {
```
mpchadwick: Wait, did I do this wrong? "app/il8n/hu" is the last element in the prior array. I'm wondering if this will screw up translations for anything after `hr`? Seems like the best things is to put `app/i18n/hu` into alphabetical order (as all the others are)?

mpchadwick: Yup it's definitely broken. Will fix this.

image

PR #359, opened 2017-11-29
Let Werkzeug create a proper response string for CORS requests

### LINKED ISSUE #297: Unicode issue in add_cors_headers with uWSGI and Python 2
While debugging some CORS issues, I found out that the 502 status I was getting back from Nginx was due to a `TypeError` exception in uWSGI:

```
  File "/usr/local/www/isso/lib/python2.7/site-packages/werkzeug/contrib/fixers.py", line 152, in __call__
    return self.app(environ, start_response)
  File "/usr/local/www/isso/lib/python2.7/site-packages/isso/wsgi.py", line 119, in __call__
    return self.app(environ, start_response)
  File "/usr/local/www/isso/lib/python2.7/site-packages/isso/wsgi.py", line 147, in __call__
    add_cors_headers("200 Ok", [("Content-Type", "text/plain")])
  File "/usr/local/www/isso/lib/python2.7/site-packages/isso/wsgi.py", line 144, in add_cors_headers
    return start_response(status, headers.to_list(), exc_info)
TypeError: http status must be a string
```

The `TypeError` gets thrown by uWSGI because the first argument to the `start_response` handler is not a string but a `unicode` object. Replacing `"200 Ok"` on line 147 with `b"200 Ok"` seems to fix the problem. I'm not a Python guru, but it looks like the problem is that all string literals are Unicode and a cursory read through the WSGI specification leads to the section [Unicode Issues](https://www.python.org/dev/peps/pep-0333/#id26), which says that all strings must be byte strings and not Unicode. uWSGI performs similar checks on the response headers, but that passes fine as Werkzeug converts the headers to latin1 byte strings in `Headers.to_list()`.

For the sake of completeness, my setup is as following:
* OS is FreeBSD 10.3-RELEASE
* Python version is 2.7.13
* `uwsgi` 2.0.14 installed from ports
* `isso` 0.10.6 from PyPI
* `Werkzeug` 0.11.15 from PyPI

Other Python WSGI apps (MoinMoin) run flawlessly.

### LINKED ISSUE #347: Isso Forces CORS Preflight but is unable to handle CORS Preflight requests
Hello!

I am attempting to use Isso in my project, but the way I am using it I have to communicate directly with the API built into Isso. I am currently running into an issue where if I attempt to add a comment using the `/new` api endpoint, the request either will return a `CSRF` error (if I omit the `Content-Type` header), or a 502 error (if I attempt to call it like I do normally).

The issue is that Isso does not respond to the CORS preflight request.

Normally, the workaround would be to preform a [Simple Request](https://developer.mozilla.org/en-US/docs/Web/HTTP/Access_control_CORS#Simple_requests) using an `XMLHttpRequest` but if the `Content-Type` header is omitted (instead of being `application/json`) so that the request is compliant with being a "Simple Request", Isso throws a [`CSRF` error](https://github.com/posativ/isso/blob/master/isso/views/comments.py#L72) since the `Content-Type` header is unset.

The way to fix this bug would be to respond properly to the CORS preflight request (by listening to an `OPTIONS` request being made to the `/new` api endpoint), or remove the requirement of adding the `Content-Type` header to allow it to be just a Simple Request.

I assume many have not run into this issue, as the CORS preflight request would not be sent if the user was on the same domain as the Isso server (since it would not be Cross Origin then).

Please let me know if you need any more information!

hiliev:
This fixes a Unicode issue in `wsgi.py` and resolves #297. It should also resolve #347.

### DIFF
--- a/isso/wsgi.py
+++ b/isso/wsgi.py
@@ -19,6 +19,7 @@ except ImportError:
 
 from werkzeug.serving import WSGIRequestHandler
 from werkzeug.wrappers import Request as _Request
+from werkzeug.wrappers import BaseResponse as Response
 from werkzeug.datastructures import Headers
 
 from isso.compat import string_types
@@ -144,8 +145,8 @@ class CORSMiddleware(object):
             return start_response(status, headers.to_list(), exc_info)
 
         if environ.get("REQUEST_METHOD") == "OPTIONS":
-            add_cors_headers("200 Ok", [("Content-Type", "text/plain")])
-            return []
+            response = Response()
+            return response(environ, add_cors_headers)
 
         return self.app(environ, add_cors_headers)
 


SnorlaxYum:
+1 for the pull request. Works fine under python 2.7 with isso, gevent and nginx.

SnorlaxYum:
The edits also works under python 3.5.3 with isso, gevent and nginx.  
I'm wondering why this haven't been merged after one year and half...... That really helps for people who have followed the doc [Advanced integration](https://posativ.org/isso/docs/extras/advanced-integration/). @jelmer

isso/wsgi.py
```diff
@@ -144,8 +145,8 @@ def add_cors_headers(status, headers, exc_info=None):
             return start_response(status, headers.to_list(), exc_info)
 
         if environ.get("REQUEST_METHOD") == "OPTIONS":
-            add_cors_headers("200 Ok", [("Content-Type", "text/plain")])
-            return []
+            response = Response()
+            return response(environ, add_cors_headers)
```
jelmer: I don't see why this needs to be changed - the issue is the first argument to werkzeug.datastructures.Headers.

jelmer: I don't see why this needs to be changed - the issue is the first argument to werkzeug.datastructures.Headers.

SnorlaxYum:
Okay, just fixed it: https://github.com/posativ/isso/pull/555 .

ix5:
Is anyone still using `uwsgi`? Otherwise we might want to remove any reference of support for it.

Closing soon unless someone can provide more information.

ix5:
I forgot that I had already looked into this in https://github.com/posativ/isso/issues/297#issuecomment-1030909525

Closing.

PR #363, opened 2017-12-15
Admin interface >> add host to comment URI

HLFH:
@blatinier Add the host to comment URI.
If ISSO URL is located here: comments.example.com
And BLOG URL is located here: example.com
And host within share/isso.conf is that: example.com
Then the comment URI was providing wrong URLs with the relative path /MY_BLOG_URL 
Now it will be: example.com/MY_BLOG_URL based on the host provided in isso.conf

### DIFF
--- a/isso/templates/admin.html
+++ b/isso/templates/admin.html
@@ -178,7 +178,7 @@ function send_edit(com_id, hash) {
       {% if order_by == "tid" %}
         {% if thread_id != comment.tid %}
           {% set thread_id = comment.tid %}
-          <h2 class="thread-title">{{comment.title}} (<a href="{{comment.uri}}">{{comment.uri}}</a>)</h2>
+          <h2 class="thread-title">{{comment.title}} (<a href="{{host}}{{comment.uri}}">{{host}}{{comment.uri}}</a>)</h2>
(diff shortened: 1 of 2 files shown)


blatinier:
@HLFH Thanks for the addition. I cannot test it right now but I'll try soon. 

Regarding the git tree, could you please squash the 3 commits into one since none of them makes sense on their own.

HLFH:
@blatinier I am more a tennis than a squash person. Thanks for the hint. Consider it done.

blatinier:
OK sorry for the review time.
There is a problem when you specify multiple hosts in the configuration. It prepend all hosts to the link, which is ugly. You should take the first one only (seems like a decent default).
As for the implementation it could be great to parse hosts in the view (split it into an array) and use only the first element in the template.
I did not saw anything else. Once you fix this it's mergeable for me.

blatinier:
Fixed with #458

pozitron57:
@blatinier I think it is not fixed with #458. In `isso/templates/admin.html`, links `<a href="{{comment.uri}}">{{comment.uri}}</a>` in my case point to `comment.domain.com/path/` instead of correct `domain.com/path/`

PR #370, opened 2018-01-13
To make the doc more specified

SnorlaxYum:
Added a systemd init script for isso standalone.
Added something important about service under the __Running Isso__ section.


SnorlaxYum:
BTW, accidentally saw this: https://comment.sh/ . 
How do u feel about it? @posativ 
In my opinion, it's not a great option when u really value the comment data. 
The only thing u can control when using it is the static js files of isso, which u can host anywhere and use.
Also, it will be the time to charge for it once they run out of the space.
It might sound great news to the ones who don't need a server, which, nowadays, everyone could obtain at a fairly reasonable price.
Sadly, with Disqus there and most of isso users treasuring their data, I don't think those people could earn much.

jelmer:
Can you add the instructions and example systemd unit file into isso git repo, rather than linking to external sites?

chrishutchison9:
@SnorlaxYum Cool you noticed my pet project. Users can always export their database under the admin. Obviously a big fan of the project and it's contributors.

PR #378, opened 2018-01-24
compat.py passes pyflakes test

cclauss:
pyflakes likes the try/except approach of the Python porting best practice of [use feature detection instead of version detection](https://docs.python.org/3/howto/pyporting.html#use-feature-detection-instead-of-version-detection).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -26,7 +26,7 @@ install:
   - sudo rm -rf /dev/shm && sudo ln -s /run/shm /dev/shm
 script:
   - tox -e $TOX_ENV
-  - python -m pyflakes.__main__ $(git ls-files | grep -E "^isso/.+.py$" | grep -v "^isso/compat.py")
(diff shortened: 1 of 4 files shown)


blatinier:
Ok this might link to your ticket about dropping 2.6 :)
If you have a simple mean to check why python2.6 fails and fix it, it could be great.
It seems that python2.6 does not raise the NameError.

But it may seal the fate of #379 at least for 2.6.
To discuss.

cclauss:
Bummer.  The __# noqa__ linter directive only works on [flake8](https://github.com/PyCQA), not on pyflakes.  The former is a superset of the latter that also adds PEP8 and code complexity tests.  The only problem that we have is that flake8 no longer supports Python 2.6.

cclauss:
Default to __memoryview__ instead of __buffer__ in all versions of Python except v2.6.

cclauss:
Too complicated with Python 2.6.  I will revisit once 2.6 is no longer supported.

PR #384, opened 2018-01-31
Upgrade from pyflakes to flake8

cclauss:

PR #386, opened 2018-01-31
autopep8 the code and then flake8

cclauss:
See #384 to understand the why...
1. run autopep8 on all Python code in the repo.  (TONS OF CHANGES!!)
2. Fix a few things by hand
3. Move from Pyflakes8 to flake8 with a short list of IGNORE issues.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,11 +13,12 @@ matrix:
       env: TOX_ENV=wheezy
 install:
   - pip install -U pip
-  - pip install pyflakes tox
+  - pip install flake8 tox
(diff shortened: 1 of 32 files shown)


jelmer:
The migration from pyflakes8 to flake8 seems reasonable to me. I'm less convinced about the mass reformatting, but perhaps others (@blatinier, @posativ) have stronger opinions on that?

blatinier:
Ok for flake8 too. The mass reformatting makes sense at least for 95% of it, the last 5% could have been formatted better to fit the pep8 but it's better this way than before.
I merge.

Issue #401, opened 2018-03-23
Apache configuration for suburl with Isso ?


Bridouz:
Hi,

I have discovered Isso yesterday and I'm fairly impressed with it. Nice piece of software.

I wanted to install it for my blog and I'm running into an issue I don't understand.

  - Isso and Blog are not on the same server
  - Isso is self-hosted on my Pi alongside Nextcloud, Gitea and Wallabag
  - I use DuckDNS to access my Pi outside
  - `myadresse.duckdns.org` leads to my Nextcloud instance
  - I want Isso to be available as `myadresse.duckdns.org/isso`

So, In Apache2 config I have : 
```
<Location "/isso">
    ProxyPass "http://localhost:8001"
    ProxyPassReverse "http://localhost:8001"
</Location>
```
My blog is able to use Isso but when a comment is submitted I receive a link to validate with `myadresse.duckdns.org/id/13/activate/MTM.DZaffA.5dORG_6UZzZRU3RcHMXuzV3NzcM`

What could I do to fix that ? (not a huge problem but still)

Thanks :smile:

blatinier:
If you are up to some code + testing + PR and take the credit for this, here is the problem:

Where the mail is constructed:
https://github.com/posativ/isso/blob/7f55c81e0875e1564eeab8ec6dd605c5699c9137/isso/ext/notifications.py#L112

Where the variable is set (probably not the good one since it's not taken from the conf, or maybe the good one but we need one more)
https://github.com/posativ/isso/blob/7f55c81e0875e1564eeab8ec6dd605c5699c9137/isso/__init__.py#L126

It uses werkzeug local context manager:
http://werkzeug.pocoo.org/docs/0.14/local/

Feel free to tackle it!

ycpei:
I just submitted a patch, in which I used the host in the general section in the conf file for the uri generation. It seems to be working in my setting in version 0.10.6 from the pypi, so can someone test this patch on the version in the master branch?

blatinier:
This won't work on some cases, for example I use isso on multiple domains, in this case `isso.conf.get("general", "host")` is a list of domains and it will result here in a `TypeError`.
You sould do some `try/except` to catch the `TypeError`and maybe take the first elements of the list or something like that…

blatinier:
Merged

Bridouz:
:+1: 

Thanks

PR #406, opened 2018-03-31
A quick dirty fix of #401

ycpei:
- Tested on version 0.10.6 on pypi, so it would be great if
someone can test it on the master version

### DIFF
--- a/isso/ext/notifications.py
+++ b/isso/ext/notifications.py
@@ -40,7 +40,8 @@ class SMTP(object):
         gh = isso.conf.get("general", "host")
         if type(gh) == str:                                                                                                                                             
             self.general_host = gh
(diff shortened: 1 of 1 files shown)


blatinier:
This modification will not work on configurations using multiple hosts. In this case the `isso.conf.get("general", "host")` holds a `list` and the concatenation operation in the mail body construction will result in a `TypeError`. You should catch it and try to fallback on the first element of the list in this case.

ycpei:
@blatinier Thanks. I have updated it to check the type of `isso.conf.get("general", "host")` before generating the link. I only have one host in general section, so couldn't test this change with multiple hosts. Let me know if it works.

ycpei:
Thanks for merging:)

PR #410, opened 2018-04-05
i18n: add space between number and CJK character

Rocket1184:
There should be space between numbers and CJK characters.
Well, I added it.


blatinier:
Thanks

PR #412, opened 2018-04-16
Provide a multi-staged Dockerfile

vincentbernat:
Many of the Docker images on hub.docker.com are outdated. The one
specified in the documentation doesn't exist anymore. We provide a
decent Dockerfile to build our own Docker image.

This uses a multi-stage build to avoid polluting the final image with
the intermediate artifacts. The final image is 155 MB. It should be
possible to squeeze it even more by using Alpine Linux for the last
two parts instead of Stretch.

The service is using gunicorn. The user is expected to complete the
installation with a reverse proxy configuration.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,33 @@
+# First, compile JS stuff
+FROM node
+WORKDIR /src/
+COPY . .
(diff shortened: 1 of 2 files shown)


blatinier:
It works great thanks

PR #418, opened 2018-04-21
html: add nofollow/noopener to links

### LINKED ISSUE #373: add rel=nofollow to links
Is there any easy way to add `rel='nofollow'` to links?

vincentbernat:
"nofollow" is a deterrent for spammers: they cannot put links and hope
to increase their SEO when all these links have the nofollow
relationship.

"noopener" is a security for links opening a new window. They ensure
the target cannot control us.

Fix #373


vincentbernat:
I didn't know what `simpletree` API look like, since it's not included in html5lib version required by Isso anymore. Maybe we could just remove this check?

PR #419, opened 2018-04-21
api: add /feed API to get an Atom feed for an URI

### LINKED ISSUE #81: RSS feed
Provide per-thread RSS (and maybe even all-comment) feeds for users that are interested in (any) follow-ups.

vincentbernat:
We need absolute URL at some places. We assume the first host
configured is the base of the URI we have.

Fix #81.

This is mostly a "RFC". Notably, tests are missing.

### DIFF
--- a/docs/docs/configuration/client.rst
+++ b/docs/docs/configuration/client.rst
@@ -20,7 +20,8 @@ preferably in the script tag which embeds the JS:
             data-isso-avatar-bg="#f0f0f0"
             data-isso-avatar-fg="#9abf88 #5698c4 #e279a3 #9163b6 ..."
             data-isso-vote="true"
-            data-vote-levels=""
+            data-isso-vote-levels=""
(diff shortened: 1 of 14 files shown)


vincentbernat:
Tests have been added. Caching should work properly and the client side too.

The only problem I see is the use of the first host defined in `[general]`. Maybe I should add an `[rss]` section instead an add a `base` parameter here instead?

blatinier:
Just like you suggested, adding a section `[rss]` in the conf for the host & limit seems like a good way to fix my comments too.
Anyway thanks for you efforts, here and on other issues, I'll try to take some time to review, test and merge all that :)

vincentbernat:
I've added a commit to include some configuration on both server-side and client-side.

vincentbernat:
Forgot to include HTML rendering. I've pushed another commit for that.

PR #438, opened 2018-05-07
feed: remove trailing / from base URL

### LINKED ISSUE #437: Comment feed: Links to comments are not working
I am running the master pulled from yesterday and so far it works nice and comment feeds are great as finally users have the possibility to stay updated but I found a malfunction.

In the [comment feed](https://blog.mdosch.de/fcgi-bin/isso/feed?uri=%2F2018%2F05%2F04%2Fseite-fuer-xmpp-einsteiger-sucht-unterstuetzer%2F) for [this site](https://blog.mdosch.de/2018/05/04/seite-fuer-xmpp-einsteiger-sucht-unterstuetzer/) you don't get to the comment if you click on a comment link as there is a wrong URL shown.

Maybe it is due to me using relative links everywhere to make my site also work as hidden service without changing links in the compiled website. Or maybe I made something wrong while [updating](https://blog.mdosch.de/2018/05/06/aktuellen-master-von-isso-auf-uberspace-6-nutzen/) as this was try and error.

vincentbernat:
This way, one can use "/" as base URL. This is only valid if we are
sure that URI should always have a leading "/". Is that the case?

Fix #437.

### DIFF
--- a/isso/views/comments.py
+++ b/isso/views/comments.py
@@ -887,7 +887,7 @@ class API(object):
         except ValueError:
             return BadRequest("limit should be integer")
         comments = self.comments.fetch(**args)
-        base = conf.get('base')
+        base = conf.get('base').rstrip('/')
         hostname = urlparse(base).netloc
 
         # Let's build an Atom feed.


craigphicks:
What cases did you test?  What external settings (e.g. nginx setup) are assumed?  What are you assuming about configuration file settings?

Issue #420, opened 2018-04-22
pull 406 breaks Activate/Delete links when subdomains are used


arno01:
Hi,

It appears that https://github.com/posativ/isso/pull/406 breaks Activate/Delete links when subdomains are used.

In my case I have isso at: comments.mysite.com
So the links coming in email are invalid:

```
Delete comment: https://mysite.com//id/123/delete/MTI...redacted
Activate comment: https://mysite.com//id/123/activate/MTI...redacted
```

> Note also there is a double slash ``//`` which should not be there.

Applying this (pretty much a revert), fixes that, so it is working well as before:

```
diff --git a/isso/ext/notifications.py b/isso/ext/notifications.py
index 2b4a81c..2900984 100644
--- a/isso/ext/notifications.py
+++ b/isso/ext/notifications.py
@@ -115,7 +115,7 @@ class SMTP(object):
                  (local("origin") + thread["uri"] + "#isso-%i" % comment["id"]))
         rv.write("\n")
 
-        uri = self.general_host + "/id/%i" % comment["id"]
+        uri = local("host") + "/id/%i" % comment["id"]
         key = self.isso.sign(comment["id"])
 
         rv.write("---\n")
```

Good links for subdomain configuration should be:

```
Delete comment: https://comments.mysite.com/id/123/delete/MTI...redacted
Activate comment: https://comments.mysite.com/id/123/activate/MTI...redacted
```

----

We still have to fix isso when its endpoint is a directory instead of a subdomain https://github.com/posativ/isso/issues/401

pellenilsson:
This hit me as well. I run Isso at a different port than the web server. The setting `isso.conf.get("general", "host")` is supposed to contain the URL of the website, not the Isso server. The setting `isso.conf.get("server", "listen")` would work in some cases, but probably not in all cases since it can be a Unix domain socket (`unix:///tmp/isso.sock`) or a local address (`http://localhost:1234/`).

I don't have the full picture of all deployment methods, so I can't really suggest a fix unfortunately. The old code (`local("host")`) only worked for uWSGI?

pellenilsson:
Hmm, in the general case I guess it's impossible for Isso to auto-detect the public URL that it can be accessed with from the outside world, since part of the configuration may be in another application (e.g. Apache or nginx). So maybe we have to introduce a new setting for that (`server.public_url` or something). If that setting is missing, it could perhaps try auto detection the old way (`local("host")`). Sounds good?

ping @ycpei

[...]

PR #458, opened 2018-07-11
Allow subdirectory server configuration, e.g. https://example.com/isso

craigphicks:
# Allow sudirectory server configuration, , e.g. https://example.com/isso

## Description

When the isso root location was 

    https://example.com/isso

the administration pages did not work correctly.  The changes in this pull request fix that and allow proper working with or without a subdirectory.  E.g. both of these become valid configurations:

    https://example.com
    https://example.com/isso

The subdirectory case only requires the subdirectory to be passed as a parameter in the proxy call.  Here is an example *nginx* setup:

    location /isso {
        proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
        proxy_set_header X-Script-Name /isso;      // <---- critical for subdirectory config
        proxy_set_header Host $host;
        proxy_set_header X-Forwarded-Proto $scheme;
        proxy_pass http://localhost:1234;
    }
 
If the line `proxy_set_header X-Script-Name /isso;` is omitted, communication with *isso* will fail.

## Testing 

Given the blog address "https:\example.com", *isso* was tested using the following sites:

    https://example.com/isso
    https://another.com
    https://another.com/isso

## Other bugs found, not fixed yet

The *active* and *delete* URLs sent by the *isso* mail notification system were found to fail under the following conditions, with the follow symptoms:

1. *Condition:* Blog and Isso servers differ.  E.g.  `https://example.com/` and `https://another.com/`.  
    *Symptom:* The *active/delete* URLs use the blog server instead of the isso server.
2. *Condition:* Isso is cofigured as a subdirectory, e.g. `https://example.com/isso`.  
    *Symptom:* The `isso` part is not included in the URL.
3. *Condition:* Both 1 and 2 together.
    *Symptom:* Both

I will try to address these mail bugs later.

### DIFF
--- a/isso/templates/admin.html
+++ b/isso/templates/admin.html
@@ -1,8 +1,8 @@
 <html>
 <head>
   <title>Isso admin</title>
-  <link type="text/css" href="/css/isso.css" rel="stylesheet">
-  <link type="text/css" href="/css/admin.css" rel="stylesheet">
+  <link type="text/css" href="{{isso_host_script}}/css/isso.css" rel="stylesheet">
+  <link type="text/css" href="{{isso_host_script}}/css/admin.css" rel="stylesheet">
 </head>
 <body>
 <script type="text/javascript">
@@ -32,16 +32,16 @@ function fade(element) {
         op -= op * 0.1;
     }, 10);
 }
-function moderate(com_id, hash, action) {
+function moderate(com_id, hash, action, isso_host_script) {
   ajax({method: "POST",
-        url: "/id/" + com_id + "/" + action + "/" + hash,
+        url: "" + isso_host_script + "/id/" + com_id + "/" + action + "/" + hash,
         success: function(){
             fade(document.getElementById("isso-" + com_id));
         }});
 }
-function edit(com_id, hash, author, email, website, comment) {
+function edit(com_id, hash, author, email, website, comment, isso_host_script) {
   ajax({method: "POST",
-        url: "/id/" + com_id + "/edit/" + hash,
+        url: "" + isso_host_script + "/id/" + com_id + "/edit/" + hash,
         data: JSON.stringify({text: comment,
                               author: author,
                               email: email,
@@ -53,11 +53,11 @@ function edit(com_id, hash, author, email, website, comment) {
           console.log("Error: ", ret); // TODO flash msg/notif
         }});
 }
-function validate_com(com_id, hash) {
-    moderate(com_id, hash, "activate");
+function validate_com(com_id, hash, isso_host_script) {
+    moderate(com_id, hash, "activate", isso_host_script);
 }
-function delete_com(com_id, hash) {
-    moderate(com_id, hash, "delete");
+function delete_com(com_id, hash, isso_host_script) {
+    moderate(com_id, hash, "delete", isso_host_script);
 }
 function unset_editable(elt_id) {
     var elt = document.getElementById(elt_id);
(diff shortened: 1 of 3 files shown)


craigphicks:
Now I have read 
https://github.com/posativ/isso/issues/420  
which proposes a configuration setting for the isso server public url.  
That would seemingly remove the need for the changes in this pull.

isso/views/comments.py
```diff
@@ -1001,7 +1001,7 @@ def login(self, env, req):
             response = redirect(re.sub(
                 r'/login$',
                 '/admin',
-                get_current_url(env, strip_querystring=True)
+                get_current_url(env, strip_querystring=True) #iwozere
```
blatinier: This comment should go away.

isso/templates/admin.html
```diff
@@ -32,16 +32,16 @@
...
+        url: "" + isso_host_script + "/id/" + com_id + "/" + action + "/" + hash,
         success: function(){
             fade(document.getElementById("isso-" + com_id));
         }});
 }
-function edit(com_id, hash, author, email, website, comment) {
+function edit(com_id, hash, author, email, website, comment, isso_host_script) {
   ajax({method: "POST",
-        url: "/id/" + com_id + "/edit/" + hash,
+        url: "" + isso_host_script + "/id/" + com_id + "/edit/" + hash,
```
blatinier: The first `"" + ` is useless

isso/templates/admin.html
```diff
@@ -32,16 +32,16 @@
         op -= op * 0.1;
     }, 10);
 }
-function moderate(com_id, hash, action) {
+function moderate(com_id, hash, action, isso_host_script) {
   ajax({method: "POST",
-        url: "/id/" + com_id + "/" + action + "/" + hash,
+        url: "" + isso_host_script + "/id/" + com_id + "/" + action + "/" + hash,
```
blatinier: The first `"" + ` is useless

craigphicks:
@blatinier - I have updated and pushed.  [This page of differences](https://github.com/posativ/isso/pull/458/files) now looks correct.

craigphicks:
Something must be added to documentation about server setting for the subdirectory case:  E.g. the below information.  That can be added in another pull if this one is ready to go.  Or it can be added in this pull.

----------------------

The subdirectory case only requires the subdirectory to be passed as a parameter in the proxy call. Here is an example nginx setup:

location /isso {
    proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
    proxy_set_header X-Script-Name /isso;      // <---- critical for subdirectory config
    proxy_set_header Host $host;
    proxy_set_header X-Forwarded-Proto $scheme;
    proxy_pass http://localhost:1234;
}

If the line proxy_set_header X-Script-Name /isso; is omitted, communication with isso will fail.

craigphicks:
I see that documentation about `proxy_set_header X-Script-Name /isso;` already exists [under docs](https://github.com/posativ/isso/blob/master/docs/docs/setup/sub-uri.rst).

craigphicks:
I have modified `isso/views/comments.py` to follow the policy started in [pull 443](https://github.com/posativ/isso/pull/443), which that the so-called 'public_endpoint' will be read from the conf file *general/public_endpoint*, and if that is not present it will be taken from the internal variable `local.host`.

This pull 458 originally introduced  `local.host` usage to to enable isso service as a subdirectory of the host URL, e.g. `https://example.com/isso`, where `/isso` is the subdirectory.  In the particular test case *isso* is proxied from an *nginx* server, with the configuration as shown above.

The purpose of the config *general/public_endpoint* setting option is as fallback in case, due to an unusual setup, or whatever.

It is worth adding that there already exists a config setting *general/name* which may have duplicate functionality with *general/public_endpoint*.  When setting up a second isso, and serving from an *nginx* location 'isso-2': 

```
location /isso-2 {
    proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
    proxy_set_header X-Script-Name /isso-2;      // <---- critical for subdirectory config
    proxy_set_header Host $host;
    proxy_set_header X-Forwarded-Proto $scheme;
    proxy_pass http://localhost:1234;
}

```
I found it was necessary to set 

```
[general]
name = isso-2
```

in order to get the connection to work.  Indeed the *isso* documentation on [Multiple Sites](https://posativ.org/isso/docs/setup/multiple-sites/) explains that it is necessary to specify the subdirectory in config under *general/name*.

Issue #449, opened 2018-06-01
Wrong URL to accept/decline moderated mails


schiessle:
I run isso under `https://mydomain.com/isso` and everything works great, but the links send to me by mail to accept/decline comments look like this `http://mydomain.com/id/...`, correct would be `https://mydomain.com/isso/id...`. Is there a way to configure the root of isso so that the links for the moderation mails are constructed the right way? I couldn't find anything in the documentation.

craigphicks:
Change the line:

        uri = self.general_host + "/id/%i" % comment["id"]

to 

        uri = local.host + "/id/%i" % comment["id"]

in *isso/ext/notifications.py*, or wait for a fixed version.

blatinier:
You can now use the `public-endpoint` configuration in the `[general]` section of isso config. It should solve your problem.
(Newly merged in master)

PR #460, opened 2018-07-12
fix mail acitvation/delete links, c.f. issue #449

craigphicks:
As mentioned by @schiessle in [issue 449](https://github.com/posativ/isso/issues/449) mail activation and delete links were not correct for case of subdirectory configuration.  (Also true for case of different server).

This fix was tested in these cases:

```
https://blogsite.com/isso
https://anothersite.com
https://anothersite.com/isso
```


blatinier:
I think this is outdated now that #443 has been merged.
Feel free to reopen if I'm wrong.

PR #466, opened 2018-07-28
add  "postbox-notification"

pozitron57:
By the way, what happens if there is no translation for some entry? Isso uses English? For example, there is no Russian translation for "atom-feed", but I think it is OK to display "Atom feed" for Russian users.


blatinier:
I must admit that I don't know this part of the project yet. So I'm not able to answer…
And to be honest if I ever go into the JS part of it I rewrite it all in some more modern way.

PR #470, opened 2018-07-30
admin: restore comment content after cancel edit

rocka:
save comment innerHTML under `elt.dataset['originContent']`, and restore it if edit action was canceled.

output


ix5:
This is a valid PR. I've pushed to https://github.com/ix5/isso/tree/abort-edit-save to use the new location of `admin.js` instead of inline js inside the template. Feel free to force-push.

rocka:
I've rebased the commit againt master branch (5f3bf4f as now), and confirmed it works on my local deployment. Could you please review it?

blatinier (approved):

blatinier:
Thanks for the contribution!

PR #482, opened 2018-09-26
use author names in gravatars

rr-:
We're using gravatars. Most of our users do not provide e-mails. With the current implementation, different people look the same (they all use default icon) and as such, the avatars aren't very useful.

The proposed change makes gravatars fall back to authors' nicknames when they didn't provide an e-mail. That way, if someone leaves multiple comments under the same nickname, they'll have the same gravatar, but they'll still be different from other users who use different nicknames.

### DIFF
--- a/isso/views/comments.py
+++ b/isso/views/comments.py
@@ -809,7 +809,7 @@ class API(object):
         if not self.conf.getboolean('gravatar'):
             return item
 
-        email = item['email'] or ""
+        email = item['email'] or item['author'] or ''
         email_md5_hash = md5(email)
 
         gravatar_url = self.conf.get('gravatar-url')
@@ -1066,7 +1066,7 @@ class API(object):
             get_current_url(env, strip_querystring=True) + '/index.html'
         )
 
-    def login(self, env, req):        
+    def login(self, env, req):
         data = req.form
         password = self.isso.conf.get("general", "admin_password")
         if data['password'] and data['password'] == password:


jelmer:
This is a bit of a hack; it sets the email variable to something that is not an email address, and makes a request to gravatar that it knows is not going to get it a proper gravatar image.

rr-:
I can put it behind a configuration variable if you want me to.

jelmer:
On Tue, Nov 13, 2018 at 07:40:15AM -0800, Marcin Kurczewski wrote:
> I can put it behind a configuration variable if you want me to.
Sorry, I should have explained better. It seems suboptimal
to send the full names of commenters to gravatar when we're not
going to get actual photos back but just generated avatars.

isso already has a mechanism for generating unique avatars based on
hashes, which we could use here, and which wouldn't leak commenter
names.

Jelmer

rr-:
>isso already has a mechanism for generating unique avatars based on
hashes, which we could use here, and which wouldn't leak commenter
names.

That's interesting, could you point me to it? Thanks

jelmer:
Isso uses the identicon library. Search the repository for that string to see how it is used.

jelmer:
ping?

rr-:
I forgot about this PR. Now that I looked at the code, this will make the site appear weird: users with e-mail but no gravatar will have one kind of default identicon, while users with no e-mail will have another kind.

jelmer:
Ah, fair enough. This is not unreasonable since it would only happen for sites that already have gravatar enabled anyway.

PR #483, opened 2018-09-27
Use bleach to sanitize HTML; allows use with newer versions of html5lib.

### LINKED ISSUE #296: fyi/ftr: html5lib dependency

jelmer:
This fixes #296

### DIFF
--- a/isso/tests/test_html.py
+++ b/isso/tests/test_html.py
@@ -63,7 +63,6 @@ class TestHTML(unittest.TestCase):
             print("Hello, World")
             </code></pre>""")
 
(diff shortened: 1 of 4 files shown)


gloomy-ghost:
I can't cherry-pick the commit with git since the bleach branch seems deleted. If you can restore the branch then I'll add some test tweaks and create a new PR.

Issue #487, opened 2018-10-10
API error message content-type ignores Accept header


sundbry:
When I hit the comment limit using the API, the error messages are coming back as HTML. I am using the API directly from custom javascript. 
```
<!DOCTYPE HTML PUBLIC "-//W3C//DTD HTML 3.2 Final//EN">
<title>403 Forbidden</title>
<h1>Forbidden</h1>
<p>3 direct responses to my-test-post</p>
```

The request headers include Content-Type: application/json and _Accept: application/json_ so the server should be responding with only JSON. A simple response like {"message": "3 direct responses to my-test-post"}  should allow clients to read the reason for the error.

ix5:
Fixed by #488

PR #488, opened 2018-10-10
Return HTTP errors in JSON format if Accept: contains application/json

sundbry:
Addresses issue #487 by returning JSON format error messages when in the request Accept: headers.

This does not affect existing client behavior as far as I can tell, from testing the /demo/index.html, it does not send Accept: application/json in the headers.

### DIFF
--- a/isso/__init__.py
+++ b/isso/__init__.py
@@ -69,7 +69,7 @@ local_manager = LocalManager([local])
 from isso import config, db, migrate, wsgi, ext, views
 from isso.core import ThreadedMixin, ProcessMixin, uWSGIMixin
 from isso.wsgi import origin, urlsplit
-from isso.utils import http, JSONRequest, html, hash
+from isso.utils import http, JSONRequest, JSONResponse, html, hash
 from isso.views import comments
 
 from isso.ext.notifications import Stdout, SMTP
@@ -81,6 +81,11 @@ logging.basicConfig(
 
 logger = logging.getLogger("isso")
 
+def error_handler(env, request, error):
+    if "application/json" in request.headers["Accept"]:
+        data = {'message': str(error)}
+        return JSONResponse(data, 500 if error.code == None else error.code)
+    return error
 
 class Isso(object):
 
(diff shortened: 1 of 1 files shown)


isso/__init__.py
```diff
@@ -81,6 +81,11 @@
 
 logger = logging.getLogger("isso")
 
+def error_handler(env, request, error):
+    if "application/json" in request.headers["Accept"]:
```
jelmer: This should be a little bit more intelligent; i.e. if the preferred encoding is not json, it should not return json.

jelmer:
This seems reasonable; one comment inline about Accept header parsing.

ix5:
@sundbry would you mind rebasing and addressing the comment by jelmer?

isso/__init__.py
ix5: What do you mean here @jelmer? `Accept-Encoding`?

ix5 (commented):
(See inline)

isso/__init__.py
jelmer: the accept header has a syntax, just doing a string search for "application/json" doesn't say whether that's actually the preferred encoding. What is the user sends "Accept: text/html, application/json" ?

See https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Accept

ix5: I see. This would be an example header:
`Accept: text/html, application/xhtml+xml, application/xml;q=0.9, image/webp, */*;q=0.8`

Perhaps create iterable by splitting on comma (+strip whitespace yada yada), filter out everything except `text/html` and `application/json`, then return based on first match, else text-only response.

ix5:
@sundbry this is a nice improvement. Please see inline comments and implement those slight tweaks when you find the time.

jelmer:
On Wed, Feb 23, 2022 at 12:51:12PM -0800, ix5 wrote:
> @ix5 commented on this pull request.
> 
> 
> 
> > @@ -81,6 +81,11 @@
>  
>  logger = logging.getLogger("isso")
>  
> +def error_handler(env, request, error):
> +    if "application/json" in request.headers["Accept"]:
> 
> I see. This would be an example header:
> `Accept: text/html, application/xhtml+xml, application/xml;q=0.9, image/webp, */*;q=0.8`
> 
> Perhaps create iterable by splitting on comma (+strip whitespace yada yada), filter out everything except `text/html` and `application/json`, then return based on first match, else text-only response.

I was hoping there was something in one of the libraries we're using that you could just reuse. Reimplementing that logic is doable but tricky - you'd also need to handle priorities and wildcards.

Jelmer

ix5:
Merged, though I had to manually override some changes. (You sadly disabled push access for maintainers to your branch, @sundbry)

PR #490, opened 2018-11-03
css: fix vertical alignment of notification checkbox

vincentbernat:
This is usually a pain to vertically align without additional markup.
The proposed solution happens to be the shortest solution I know of,
but it may be outdated.

The second commit is to reduce the size of the label. I have put it in another commit as it may not be approved. Feel free to drop it.


jelmer:
Could you post a screenshot with what things look like with the proposed changes?

vincentbernat:
Before the change:

image

After the change:

image

jelmer:
Thanks! That's quite a subtle changes, but seems reasonable to me.

PR #499, opened 2018-11-11
Don't open the admin page with a default password by default. Fixes #491

### LINKED ISSUE #491: isso 0.11 default install exposes admin interface with a default password: "please_choose_a_strong_password"
This seems like a security problem, anyone upgrading from 0.10 will be surprised I'm sure. I only figured out the existence of this admin interface by poking around pull requests.

This seems like something that should be disabled by default. If not, atleast documented in the public docs at [Documentation](https://posativ.org/isso/docs/). One of the things that drew me to isso was the fantastic documentation and very good defaults that prevent shooting in the foot type scenarios.

jelmer:
Don't open the admin page with a default password by default. Fixes #491

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,6 +4,9 @@ Changelog for Isso
 0.11.2 (unreleased)
 -------------------
 
+- Don't enable admin interface with default password by default.
+  (Jelmer Vernooĳ, #491)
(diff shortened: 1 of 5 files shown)


jelmer:
Ping @blatinier

PR #511, opened 2018-12-15
Consistency and link to client config

fliiiix:
Idk maybe it would be possible to link it relativ but I didn't found out how it would work.

### DIFF
--- a/docs/docs/configuration/server.rst
+++ b/docs/docs/configuration/server.rst
@@ -262,19 +262,19 @@ reply-to-self
     the comment. After the editing timeframe is gone, commenters can reply to
     their own comments anyways.
 
-    Do not forget to configure the client.
+    Do not forget to configure the `client <client>`_ accordingly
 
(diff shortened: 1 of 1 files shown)


jelmer:
Please don't use absolute links - it breaks offline access to the docs. You may want to create references to client.html instead to link to the local file?

fliiiix:
fair point - I will try

fliiiix:
This should work.

jelmer:
This is still using an absolute path; on my system it links to /configuration/client (i.e. on the filesystem root)

fliiiix:
It's absolute to the root of your webserver. Which worked for the render thing I tried. How do you use it?

Can you tell me with what you are rendering the rst files?

jelmer:
That assumes the docs are located at the root, it i.e. breaks when I access a local copy of the docs.

fliiiix:
How about this?

PR #514, opened 2018-12-22
Override translations

brad:
Hey guys,
I don't know if this has been discussed before, but my designer didn't particular like all the wording so I made this change to enable overriding translations. It's working well for me so far.

### DIFF
--- a/docs/docs/configuration/client.rst
+++ b/docs/docs/configuration/client.rst
@@ -35,7 +35,7 @@ Additionally, you can override any translation string for any language by adding
 a `data-isso-` attribute that is equal to the translation key (found `here`__) with
 `-text-[lang]` appended to it. So, for example, if you want to override the
 english translation of the `postbox-notification` message, you could add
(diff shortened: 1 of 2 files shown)


jelmer:
I'm not sure about this; couldn't we rather just fix the translations?

brad:
@jelmer 🤔 I don't know. It's a valid translation, just not exactly what she wanted. The translation in question is the `postbox-notification` checkbox. She wanted to change `Subscribe to email notification of replies` to `Select to be notified of replies to your comment`

brad:
@jelmer Please don't hesitate to close this is if it's not something you want. I won't take offense and I'm perfectly content to use continue using this in my fork.

jelmer:
I'm not sure about it; I'm hoping one of the other maintainers has a stronger opinion for or against.

I'm on the fence because more options make things harder to configure, and it's harder they'll regress because they're not tested.

blatinier:
Nice and simple implementation not so error prone and I can't say I'm against since I changed translations in every installation of isso I did :)

PR #520, opened 2019-01-02
Enable the generic migrate function

matolivier:
Greetings,

The script migrate.py is containing a tool for importing a generic format of comments.
But until now this tool was unreachable from isso binary.

These commits enable the usage of "generic" type of importation. Also I have added some documentation about how using it. The documentation part looks big, I am open to any simplifications.

Regards,
Matthieu

### DIFF
--- a/docs/docs/quickstart.rst
+++ b/docs/docs/quickstart.rst
@@ -62,7 +62,7 @@ Migration
 
 Isso provides a tool for importing comments from Disqus_ or WordPress_.
 You can also import comments form any other comment system, but this topic is more
-complex and covered in :doc:`advanced migration <extra/advanced-migration>`.
+complex and covered in :doc:`advanced migration <extras/advanced-migration>`.
 
 To export your comments from Disqus, log into Disqus, go to your website, click
 on *Discussions* and select the *Export* tab. You'll receive an email with your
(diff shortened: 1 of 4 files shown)


docs/docs/quickstart.rst
```diff
@@ -70,15 +71,39 @@ To export comments from your previous WordPress installation, go to *Tools*,
 export your data. It has been reported that WordPress may generate broken XML.
 Try to repair the file using ``xmllint`` before you continue with the import.
 
-Now import the XML dump:
+For any other comment system you can use the generic JSON format. It's up to you
```
jelmer: Can you perhaps move this to a separate file and link to it from the quickstart?

jelmer:
Thanks for working on this. Just one minor comment - I think it makes sense to keep the quickstart short.

docs/docs/quickstart.rst
matolivier: I could create a new page about importing json files in docs/extra and add a link from the quickstart.

What do you think?

jelmer: That seems like a good idea

matolivier:
Now the documentation is separated.

Issue #523, opened 2019-01-19
isso won't start after upgrading to 0.12.0


captainark:
Hi,

I can't get isso to start after upgrading to the 0.12.0 version.

I'm getting the following error message in the logs :
```
Jan 19 08:36:51 dipan isso[1912]: Error: no object bound to host
```

I'm using gunicorn, and starting isso with the following systemd unit file :
```
[Unit]
Description=Lightweight Disqus alternative

[Service]
Type=simple
User=isso
Group=isso
SyslogIdentifier=isso
WorkingDirectory=/opt/isso
Environment="ISSO_SETTINGS=/etc/isso/isso.conf"
ExecStart=/opt/isso/bin/gunicorn -b 0.0.0.0:8080 -w 2 --preload isso.run
ExecStartPost=/bin/bash -c 'echo $MAINPID > /run/isso/isso.pid'
PIDFile=/run/isso/isso.pid

[Install]
WantedBy=multi-user.target
```

Here's my configuration :
```
[general]
dbpath = /opt/isso/isso.db
name =
host =
  http://www.mywebsite.net
  https://www.mywebsite.net
max-age = 15m
notify = smtp
reply-notifications = false
gravatar = true
gravatar-url = https://www.gravatar.com/avatar/{}?d=identicon

[admin]
enabled = true
password = redacted

[moderation]
enabled = false
purge-after = 30d

[server]
listen = http://0.0.0.0:8080
public-endpoint =
reload = off
profile = off

[smtp]
username = isso
password =
host = localhost
port = 25
security = none
to = me@me.com
from = you@you.net
timeout = 10

[guard]
enabled = true
ratelimit = 5
direct-reply = 10
reply-to-self = false
require-author = false
require-email = false

[markup]
options = strikethrough, autolink, fenced_code, no_intra_emphasis
allowed-elements =
allowed-attributes =

[hash]
salt = redacted
algorithm = pbkdf2

[rss]
base =
limit = 100
```

jelmer:
Looks like a regression from 1a3f803b966cd32da367978f306b923d433c5cda

captainark:
0.12.1 is working with my configuration.

Thanks !

HorlogeSkynet:
Oups sorry guys, it looks like I missed this one... 'will try to come up with a proper fix in the future. Bye 👋

HorlogeSkynet:
@captainark #527 should not break with your configuration.
[...]

PR #527, opened 2019-01-25
[PATCH v2] Removes potential '//' from email links (#420, #519, #523)

HorlogeSkynet:
Re-submission of a non-regressing patch for a #420 detail, after #523.
The one-liner has been broken for readability purposes.

### DIFF
--- a/isso/ext/notifications.py
+++ b/isso/ext/notifications.py
@@ -79,8 +79,7 @@ class SMTP(object):
         self.public_endpoint = isso.conf.get("server", "public-endpoint")
         # rstrips potential trailing '/', without having to `repr` the `local` object.
         if self.public_endpoint:
-            if self.public_endpoint.endswith('/'):
-                self.public_endpoint = self.public_endpoint.rstrip('/')
+            self.public_endpoint = self.public_endpoint.rstrip('/')
         else:
             self.public_endpoint = local("host")
 
--- a/isso/ext/notifications.py
+++ b/isso/ext/notifications.py
@@ -76,7 +76,14 @@ class SMTP(object):
 
         self.isso = isso
         self.conf = isso.conf.section("smtp")
-        self.public_endpoint = isso.conf.get("server", "public-endpoint") or local("host")
+        self.public_endpoint = isso.conf.get("server", "public-endpoint")
+        # rstrips potential trailing '/', without having to `repr` the `local` object.
+        if self.public_endpoint:
+            if self.public_endpoint.endswith('/'):
+                self.public_endpoint = self.public_endpoint.rstrip('/')
+        else:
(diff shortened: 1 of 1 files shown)


HorlogeSkynet:
Hey over here, could this one be merged ?
Thanks ! Bye :wave:

arno01:
I've tested it, it is working for me.
@blatinier @jelmer

isso/ext/notifications.py
```diff
@@ -76,7 +76,14 @@ def __init__(self, isso):
 
         self.isso = isso
         self.conf = isso.conf.section("smtp")
-        self.public_endpoint = isso.conf.get("server", "public-endpoint") or local("host")
+        self.public_endpoint = isso.conf.get("server", "public-endpoint")
+        # rstrips potential trailing '/', without having to `repr` the `local` object.
```
jelmer: I don't understand this comment - what does repr have to do with it? Why is it bad to strip / from the output of local("host") ?

isso/ext/notifications.py
```diff
@@ -76,7 +76,14 @@ def __init__(self, isso):
...
+        if self.public_endpoint:
+            if self.public_endpoint.endswith('/'):
```
jelmer: Ask for forgiveness, not permission: just try to rstrip('/') rather than checking if there is a trailing slash first.

isso/ext/notifications.py
HorlogeSkynet: Please see #523. The first patch (https://github.com/posativ/isso/commit/1a3f803b966cd32da367978f306b923d433c5cda) dealt with the output of `local("host")` as a string and hard-failed **at this step**.
That's why the r-strip is only applied on an URL fetched from configuration.

HorlogeSkynet:
Up @jelmer 👍

jelmer:
Sorry for the delay in reviewing, I finally spend some time trying to understand what's been broken in the various pull requests.

isso/ext/notifications.py
```diff
@@ -76,7 +76,13 @@ def __init__(self, isso):
 
         self.isso = isso
         self.conf = isso.conf.section("smtp")
-        self.public_endpoint = isso.conf.get("server", "public-endpoint") or local("host")
+        self.public_endpoint = isso.conf.get("server", "public-endpoint")
+        # rstrips potential trailing '/', without having to `repr` the `local` object.
+        if self.public_endpoint:
+            self.public_endpoint = self.public_endpoint.rstrip('/')
+        else:
+            self.public_endpoint = local("host")
```
ix5, line 84: This is redundant. Other places will always use `isso_host_script = conf.get("server", "public-endpoint") or local.host` anyway.

PR #529, opened 2019-01-28
contrib: Add Blogger importer tool

mardy:
With this tool I could convert all the comments from blogger.com into a JSON file which then I imported into isso. Here's the result:

http://www.mardy.it/ia/blog/2018/11/and-now-for-something-completely.html


mardy:
That was my initial thought, but I decided against it for a couple of reasons:

1) in this way, one can easily preview the JSON file before importing the comments;
2) the code quality of this script is sub-optimal, I just wanted to quickly reach my goal and share it with others. I'm afraid I don't have time to polish it, because I would also need to setup a testing environment in order not to mess up with my real blog.

I'm definitely willing to fix any review comments, and also help with reviews if someone wants to improve this (or integrate it into `isso/migrate.py`), but I'm afraid I cannot commit to more than that. :-(

jelmer:
Can you add some comments to the script, at least some instructions on how to use it? I'm not a big fan of contrib directories (since the files there tend to go stale and bit rot), but I guess bundling something (while not as useful as adding support in migrate.py) is at least better than nothing.

mardy:
> Can you add some comments to the script, at least some instructions on how to use it?

Done. I've also added a second commit with very minor changes to the documentation.

PR #534, opened 2019-02-15
Implement #358 (Redirect to comment after moderation)

### LINKED ISSUE #358: Make activate comment link take you directly to the comment
Hi, what do you think about this:

When clicking activate or delete, there would be no confirmation, but it would take you directly to the right comment, so you can reply.

Currently I have to go multiple times between the e-mail and the browser:
1. receive e-mail
1. click activate
1. click activate on the confirmation dialog
1. go back to the e-mail
1. click on the comment link
1. reply

It would be nicer like this:
1. receive e-mail
1. click activate
1. reply

Kwpolska:
This closes #358. After a comment is successfully moderated, the user is redirected to the comment page itself.

### DIFF
--- a/isso/views/comments.py
+++ b/isso/views/comments.py
@@ -6,6 +6,7 @@ import re
 import cgi
 import time
 import functools
+import json  # json.dumps to put URL in <script>
 
 from datetime import datetime, timedelta
 from itsdangerous import SignatureExpired, BadSignature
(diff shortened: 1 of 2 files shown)


isso/views/comments.py
```diff
@@ -516,14 +517,13 @@ def delete(self, environ, request, id, key=None):
...
         &lt;html&gt;
-            &lt;head&gt;
-                &lt;script&gt;
-                    if (confirm('Delete: Are you sure?')) {
-                        xhr = new XMLHttpRequest;
-                        xhr.open('POST', window.location.href);
-                        xhr.send(null);
-                    }
-                &lt;/script&gt;
+        &lt;head&gt;
```
jelmer: This change appears unrelated, why is it necessary?

Kwpolska: I updated some docs on occasion.

jelmer: Please split that change out to a separate pull request.

Kwpolska: Undone. Feel free to reintroduce yourself, I’m not going to create a PR for this.

jelmer: Thanks; it's just that random unrelated changes make the pull request and subsequent merge commit harder to understand, especially for somebody not intimately familiar with all parts of isso's codebase, like myself.

Kwpolska:
Thanks for merging!

Issue #306, opened 2017-03-02
BUG: Uncaught TypeError: Cannot set property 'textContent' of null


zaygraveyard:
### Isso version
`0.10.6`

### How to reproduce

1. Load a page with comments
2. Remove the `#isso-thread` element (`document.getElementById('isso-thread').remove()`)
3. Wait a minute

### Expected

No errors

### Actual

Uncaught TypeError: Cannot set property 'textContent' of null

### Cause

In [this line](https://github.com/posativ/isso/blob/master/isso/js/app/isso.js#L137) the `$(...)` returns `null` since the element has been removed.

### Solution

```js
var timeNode = $(".permalink > time", el);
if (timeNode) {
    timeNode.textContent = utils.ago(
        globals.offset.localTime(), new Date(parseInt(comment.created, 10) * 1000));
    setTimeout(refresh, 60*1000);
}
```

### Why am I removing the element?

I have the comments component in one of the pages of an SPA (Single Page App).
I hacked the `embed.js` script to allow the building of the component on demand (not just on load).

ix5:
Isso has had changes to its client code, could you verify this is still current @zaygraveyard ?

zaygraveyard:
It partially fixes it but the problem is still there because when the page contains no elements with id `isso-root`, `$('#isso-root')` returns `null` which then causes `$('#isso-root').length` to throw a `TypeError: can't access property "length", $(...) is null`.

https://github.com/posativ/isso/blob/e6dbd21819f8fdca9fdace447e9ba01276e8d6a9/isso/js/embed.js#L51-L53

The correct fix should probably be:
```js
 if (!$('#isso-root')) { 
     return; 
 } 
```

FWIW, looking at the implementation of `$()`, `$().length` should never be used because `$()` could return `null`, an `Element` or an `Array` and since `null` and `Element`s don't have a `length` property.

ix5:
@zaygraveyard While I am not able to reproduce your issue with my current installation of isso (via `document.getElementById('isso-thread').remove()`), if this is important to you, feel free to submit a pull request for the fix to the `refresh()` function and to the check for `#isso-root`.

You might also want to change the way the timers are created. Right now, for every comment a separate self-refreshing timer  is created (unnecessary waste of cycles and might get throttled by future browsers). My idea would be to have one global timer that updates all the comment human-readable times at once.

zaygraveyard:
@ix5 I submitted a PR #757 fixing the mentioned issue.

[...]

PR #542, opened 2019-03-05
Add slovak translations + fix TypeError in embed.js

### LINKED ISSUE #372: Uncaught TypeError in embed.js
image

`Uncaught TypeError: $(...).append is not a function`

Code around the error:

    domready(function() {

        if (config["css"]) {
            var style = $.new("style");
            style.type = "text/css";
            style.textContent = css.inline;
            $("head").append(style);
        }

        count();

        if ($("#isso-thread") === null) {
            return console.log("abort, #isso-thread is missing");
        }

        $("#isso-thread").append($.new('h4')); // This line
        $("#isso-thread").append(new isso.Postbox(null));
        $("#isso-thread").append('<div id="isso-root"></div>');

        api.fetch($("#isso-thread").getAttribute("data-isso-id"),
            config["max-comments-top"],
            config["max-comments-nested"]).then(
            function(rv) {
                if (rv.total_replies === 0) {
                    $("#isso-thread > h4").textContent = i18n.translate("no-comments");
                    return;
                }

It's the newest version in pip source.

krtcom:
Add slovak translations
Stop fetching comments if '#isso-thread' does not exists, should fix #372 and #306

### DIFF
--- a/docs/docs/extras/advanced-migration.rst
+++ b/docs/docs/extras/advanced-migration.rst
@@ -4,12 +4,10 @@ Advanced Migration
 In quickstart we saw you can import comments from Disqus or WordPress. But there
 are a many other comments system and you could be using one of them.
 
-Isso provides a way to import such comments, however it's up to you to :
+Isso provides a way to import such comments, however it's up to you to to:
 
 - dump comments
-- fit the data to the following JSON format
-
-.. code-block::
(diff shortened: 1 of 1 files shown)


jelmer:
Thanks for contributing! Please split these two changes into separate pull requests, as they're unrelated.

PR #544, opened 2019-03-11
cancel fetching comments if #isso-root is not available

krtcom:


krtcom:
Stop fetching comments if '#isso-thread' does not exists, should fix #372  and #306 
As requested in pull request #542 I split this into two pull requests

PR #518, opened 2018-12-29
Email Notification Customization

SnorlaxYum:
## Mail Format Choose
Available options: `plain`, `html`, `multipart`. Default set to `plain`, so you don't need to config this in the mail block of the isso server conf if you mean to use plain text:

	[mail]
	...
	format = plain

To use multipart, set the option to `multipart` in the mail block of the isso server conf:

	[mail]
	...
	format = multipart

To use html, set the option to `html` in the mail block of the isso server conf:

	[mail]
	...
	format = html

## Title Customization  

To customize the mail title, set like the following options in the mail block of the isso server conf:  

	[mail]
	...
	title_admin = {replier} commented on {title}
	title_user = {receiver}, {replier} replied your comments on {title}

**Result:**  

## Mail Language

	[mail]
	...
	language = en

https://github.com/posativ/isso/pull/518#issuecomment-450546330

## Mail Customization  

All three formats support customization, it helps to read the dafault configuration for details (`isso/templates/comment.html` for `format = html`, `isso/templates/comment.plain` for `format = plain`).

To use the feature, set the following options like this in the mail block of the isso server conf:   

	[mail]
	...
	template = /the/path/to/the/directory/or/a-single-template

If you set it to the path to a directory, 
for `format = html` it will check if `admin.html` and `user.html` in that directory are available. 
(For `format = plain` it will check `admin.plain` and `user.plain`; 
For `format = multipart`, it will check `admin.html` and `user.html` for html part, `admin.plain` and `user.plain` for plain part) 
The name of the file should be self-explaining. `user.html` and `user.plain` are used for reply notifications to the subcribed users, `admin.html` and `admin.plain` are used for mail notifications to the admin.  
If one of these is missing then in the corresponding case the template will fall back to the default.
The log will tell you what happened when the error occurs. The template will be rendered by Jinja2.

Alternatively, you could just directly edit the default mail template without setting the option `template`.

Default template path for `format = plain` (Different if you set a different mail language using the option `language`):  
When mail language is not set using `language` or set via `language = en`: `isso/templates/comment.plain`
When it's set any other language  than `en` using `language`, take `language = ja` for example:  `isso/templates/comment_ja.plain`

Default template path for `format = html` (Different if you set a different mail language using the option `language`):  
When mail language is not set using `language` or set via `language = en`: `isso/templates/comment.html`
When it's set any other language  than `en` using `language`, take `language = ja` for example:  `isso/templates/comment_ja.html`

Default template path for `format = multipart`:  
[...]

PR #553, opened 2019-05-29
Mail subject customization

SnorlaxYum:
## Subject Customization  

To customize the mail title, set like the following options in the mail block of the isso server conf:  

	[mail]
	...
	subject_admin = {replier} commented on {title}
	subject_user_new_comment = {receiver}, {replier} replied {repliee}'s comments on {title}
	subject_user_reply = {replier} replied to your comments on {title}

In the example, `subject_user` is set to two values, then the first one will be used when {repliee}'s comment is not the {receiver}'s original one, the second one will be used otherwise. 
`subject_user` can be set to a value, then the value will be used all the time in the reply notification.

**Result:**  

I'll open pull requests about new features to the newly created mail block once the PR is merged. If you wanna test, you can check [it](https://github.com/posativ/isso/pull/518) out in advance.

### DIFF
--- a/docs/docs/configuration/server.rst
+++ b/docs/docs/configuration/server.rst
@@ -229,6 +229,35 @@ timeout
     specify a timeout in seconds for blocking operations like the
     connection attempt.
 
+Mail
+----
+
+configure the format the mail notification uses in SMTP section.
+
+.. code-block:: ini
+
+    [mail]
+    subject_admin = {title}
+    subject_user = Re: New comment posted on {title}
+
+subject_admin
+  	specify the subject format of the notification email sent to the admin.
+    Default: `{title}`
+    Available variables:
+    `{title}`: The title of the isso thread.
+    `{replier}`: The name of the author of the comment. If the author didn't fill in a
+    name, it will fallback to the term "Anonymous".
+
+subject_user
+  	specify the subject format of the notification email sent to the subscribed
+  	commenter.
+    Default: `Re: New comment posted on {title}`
+    Available variables:
+    `{title}`: The title of the isso thread.
+    `{replier}`: The name of the author of the comment. If the author didn't fill in a
+    name, it will fallback to the term "Anonymous".
+    `{receiver}`: The author whose comment is replied to in this situation.  If the
+    author didn't fill in a name, it will fallback to to the term "Anonymous".
 
 Guard
 -----
--- a/docs/docs/configuration/server.rst
+++ b/docs/docs/configuration/server.rst
@@ -241,27 +241,42 @@ Configure the format the mail notification uses in the SMTP section.
     subject_user = Re: New comment posted on {title}
 
 subject_admin
-  	specify the subject format of the notification email sent to the admin.
+    specify the subject format of the notification email sent to the admin.
+
     Default: `{title}`
+
     Available variables:
-    `{title}`: The title of the isso thread.
-    `{replier}`: The name of the author of the comment. If the author has no name,
-    it will fallback to the term "Anonymous".
+
+    {title}
+        The title of the isso thread.
+
+    {replier}
+        The name of the author of the comment. If the author has no name,
+        it will fallback to the term "Anonymous".
 
 subject_user
-  	specify the subject format of the notification email sent to the subscribed
-  	commenter.
+    specify the subject format of the notification email sent to the subscribed
+    commenter. If it is set to two values, then the first one will be used when
+    `{repliee}`'s comment is not the `{receiver}`'s original one, the second one
+    will be used otherwise.
+
     Default: `Re: New comment posted on {title}`
+
     Available variables:
-    `{title}`: The title of the isso thread.
-    `{replier}`: The name of the author of the comment. If the author has no name,
-    it will fallback to the term "Anonymous".
-    `{repliee}`: The author whose comment is replied to.  If the author has no
-    name, it will fallback to the term "Anonymous".
-    `{receiver}`: The receiver of the email.
-    If it is set to two values, then the first one will be used when `{repliee}`'s
-    comment is not the `{receiver}`'s original one, the second one will be used
-    otherwise.
+
+    {title}
+        The title of the isso thread.
+
+    {replier}
+        The name of the author of the comment. If the author has no name,
+        it will fallback to the term "Anonymous".
+
+    {repliee}
+        The author whose comment is replied to.  If the author has no
+        name, it will fallback to the term "Anonymous".
+
+    {receiver}
+        The receiver of the email.
 
 Guard
 -----
--- a/docs/docs/configuration/server.rst
+++ b/docs/docs/configuration/server.rst
@@ -256,8 +256,9 @@ subject_user
     `{title}`: The title of the isso thread.
     `{replier}`: The name of the author of the comment. If the author has no name,
     it will fallback to the term "Anonymous".
-    `{receiver}`: The author whose comment is replied to in this situation.  If the
-    author has no name, it will fallback to the term "Anonymous".
+    `{repliee}`: The author whose comment is replied to.  If the author has no
+    name, it will fallback to the term "Anonymous".
+    `{receiver}`: The receiver of the email.
 
 Guard
 -----
--- a/docs/docs/configuration/server.rst
+++ b/docs/docs/configuration/server.rst
@@ -252,11 +252,30 @@ subject_admin
         The name of the author of the comment. If the author has no name,
         it will fallback to the term "Anonymous".
 
-subject_user
+subject_user_reply
     specify the subject format of the notification email sent to the subscribed
-    commenter. If it is set to two values, then the first one will be used when
-    ``{repliee}``'s comment is not the ``{receiver}``'s original one, the second one
-    will be used otherwise.
+    commenter when the new comment is a reply to his comment.
+    Default: ``Re: New comment posted on {title}``.
+    Available variables:
+
+    {title}
+        The title of the isso thread.
+
+    {replier}
+        The name of the author of the comment. If the author has no name,
+        it will fallback to the term "Anonymous".
+
+    {repliee}
+        The author whose comment is replied to.  If the author has no
+        name, it will fallback to the term "Anonymous".
+
+    {receiver}
+        The name of the recipient of the email.
+
+subject_user_new_comment
+    specify the subject format of the notification email sent to the subscribed
+    commenter whose comment is a reply to a comment when the new comment is also
+    a reply to that comment.
     Default: ``Re: New comment posted on {title}``.
     Available variables:
 
--- a/docs/docs/configuration/server.rst
+++ b/docs/docs/configuration/server.rst
@@ -259,6 +259,9 @@ subject_user
     `{repliee}`: The author whose comment is replied to.  If the author has no
     name, it will fallback to the term "Anonymous".
     `{receiver}`: The receiver of the email.
+    If it is set to a list of two strings, then the first one will be used when
+    `{repliee}`'s comment is not the `{receiver}`'s original one, the second one will be used
+    otherwise.
 
 Guard
 -----
--- a/docs/docs/configuration/server.rst
+++ b/docs/docs/configuration/server.rst
@@ -242,7 +242,7 @@ Configure the format the mail notification uses in the SMTP section.
 
 subject_admin
     specify the subject format of the notification email sent to the admin.
-    Default: ``{title}``. 
+    Default: ``{title}``.
     Available variables:
 
     {title}
@@ -272,7 +272,7 @@ subject_user
         name, it will fallback to the term "Anonymous".
 
     {receiver}
-        The receiver of the email.
+        The name of the recipient of the email.
 
 Guard
 -----
--- a/docs/docs/configuration/server.rst
+++ b/docs/docs/configuration/server.rst
@@ -259,8 +259,8 @@ subject_user
     `{repliee}`: The author whose comment is replied to.  If the author has no
     name, it will fallback to the term "Anonymous".
     `{receiver}`: The receiver of the email.
-    If it is set to a list of two strings, then the first one will be used when
-    `{repliee}`'s comment is not the `{receiver}`'s original one, the second one will be used
+    If it is set to two values, then the first one will be used when `{repliee}`'s
+    comment is not the `{receiver}`'s original one, the second one will be used
     otherwise.
 
 Guard
--- a/docs/docs/configuration/server.rst
+++ b/docs/docs/configuration/server.rst
@@ -238,7 +238,8 @@ Configure the format the mail notification uses in the SMTP section.
 
     [mail]
     subject_admin = {title}
-    subject_user = Re: New comment posted on {title}
+    subject_user_reply = Re: New comment posted on {title}
+    subject_user_new_comment = Re: New comment posted on {title}
 
 subject_admin
     specify the subject format of the notification email sent to the admin.
--- a/docs/docs/configuration/server.rst
+++ b/docs/docs/configuration/server.rst
@@ -232,7 +232,7 @@ timeout
 Mail
 ----
 
-configure the format the mail notification uses in SMTP section.
+Configure the format the mail notification uses in the SMTP section.
 
 .. code-block:: ini
 
@@ -245,8 +245,8 @@ subject_admin
     Default: `{title}`
     Available variables:
     `{title}`: The title of the isso thread.
-    `{replier}`: The name of the author of the comment. If the author didn't fill in a
-    name, it will fallback to the term "Anonymous".
+    `{replier}`: The name of the author of the comment. If the author has no name,
+    it will fallback to the term "Anonymous".
 
 subject_user
   	specify the subject format of the notification email sent to the subscribed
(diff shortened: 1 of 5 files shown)


isso/ext/notifications.py
```diff
@@ -148,8 +148,11 @@ def format(self, thread, comment, parent_comment, recipient=None, admin=False):
 
     def notify_new(self, thread, comment):
         if self.admin_notify:
+            subject = self.isso.conf.get("mail", "subject_admin").format(
+                title=thread["title"],
+                replier=comment["author"] or self.no_name)
```
jelmer: Where does self.no_name get set?

docs/docs/configuration/server.rst
```diff
@@ -229,6 +229,35 @@ timeout
     specify a timeout in seconds for blocking operations like the
     connection attempt.
 
+Mail
+----
+
+configure the format the mail notification uses in SMTP section.
```
jelmer: Please capitalize the first word.

s/uses in SMTP section/uses in the SMTP section/

docs/docs/configuration/server.rst
```diff
@@ -229,6 +229,35 @@ timeout
...
+subject_user
+  	specify the subject format of the notification email sent to the subscribed
+  	commenter.
+    Default: `Re: New comment posted on {title}`
+    Available variables:
+    `{title}`: The title of the isso thread.
+    `{replier}`: The name of the author of the comment. If the author didn't fill in a
+    name, it will fallback to the term "Anonymous".
+    `{receiver}`: The author whose comment is replied to in this situation.  If the
+    author didn't fill in a name, it will fallback to to the term "Anonymous".
```
jelmer: Double "to"

share/isso.conf
```diff
@@ -134,6 +134,28 @@ from =
...
+# {title}: The title of the isso thread.
+# {replier}: The name of the author of the comment. If the author didn't fill in a Name
+# blank, it will fallback to the term "Anonymous".
+subject_admin = {title}
+
+# specify the subject format of the notification email sent to the subscribed
+# commenter. Default: Re: New comment posted on {title}
+# Available variables:
+# {title}: The title of the isso thread.
+# {replier}: The name of the author of the comment. If the author didn't fill in a Name
```
jelmer: "didn't fill in a Name blank" ?

SnorlaxYum:
Okay, any feedback?

isso/ext/notifications.py
```diff
@@ -79,13 +80,17 @@ def __init__(self, isso):
...
         self.reply_notify = isso.conf.getboolean("general", "reply-notifications")
+        self.no_name = "Anonymous"
 
         # test SMTP connectivity
         try:
             with SMTPConnection(self.conf):
                 logger.info("connected to SMTP server")
         except (socket.error, smtplib.SMTPException):
-            logger.exception("unable to connect to SMTP server")
+            if os.path.join(dist.location, "isso", "tests", "test_mail.py") in sys.argv:
```
jelmer: Please don't special case behaviour for tests.

isso/ext/notifications.py
```diff
@@ -79,13 +80,17 @@ def __init__(self, isso):
         self.public_endpoint = isso.conf.get("server", "public-endpoint") or local("host")
         self.admin_notify = any((n in ("smtp", "SMTP")) for n in isso.conf.getlist("general", "notify"))
         self.reply_notify = isso.conf.getboolean("general", "reply-notifications")
+        self.no_name = "Anonymous"
```
jelmer: This should be a constant, not an instance variable.

isso/ext/notifications.py
```diff
@@ -104,9 +109,10 @@ def __iter__(self):
         yield "comments.new:after-save", self.notify_new
         yield "comments.activate", self.notify_activated
 
-    def format(self, thread, comment, parent_comment, recipient=None, admin=False):
+    def format(self, thread, comment, parent_comment, recipient=None):
 
         rv = io.StringIO()
+        admin = not parent_comment
```
jelmer: I don't understand this logic, why are we in admin mode if there is no parent comment?

isso/ext/notifications.py
```diff
@@ -104,9 +109,10 @@ def __iter__(self):
         yield "comments.new:after-save", self.notify_new
         yield "comments.activate", self.notify_activated
 
-    def format(self, thread, comment, parent_comment, recipient=None, admin=False):
+    def format(self, thread, comment, parent_comment, recipient=None):
```
jelmer: Why is this changing the signature of format? It should only be touching upon the subject.

isso/ext/notifications.py
```diff
@@ -139,17 +145,43 @@ def format(self, thread, comment, parent_comment, recipient=None, admin=False):
...
+            key = self.isso.sign(('unsubscribe', recipient["email"]))
 
-            rv.write("Unsubscribe from this conversation: %s\n" % (uri + "/unsubscribe/" + quote(recipient) + "/" + key))
+            rv.write("Unsubscribe from this conversation: %s\n" % (uri + "/unsubscribe/" + quote(recipient["email"]) + "/" + key))
 
         rv.seek(0)
         return rv.read()
 
+    def notify_subject(self, thread, comment, parent_comment=None, recipient=None):
+        if parent_comment:
```
jelmer: Again, why does parent_comment=None imply admin?

SnorlaxYum:
Comment 1: But the test should not ask for SMTP credentials, that's the best way I could think of to pass the smtp test.
Comment 2: Okay, this variable will be translated according to `lang` in the `mail` block. (Will be added in my future PR). But yes, that's not needed in this PR for now.  
Comment 3~5: 
Okay, in the original version, here's the code about how it calls the `format` function before sending mail to admin:  

    def notify_new(self, thread, comment):
        if self.admin_notify:
            body = self.format(thread, comment, None, admin=True)
            self.sendmail(thread["title"], body, thread, comment)

        if comment["mode"] == 1:
            self.notify_users(thread, comment)

It assgigns `None` to the parameter `parent_comment` in this case, 'cause to notify admin of the new comment, `parent_comment` is not needed.  

How it calls the `format` function before sending mail to subcribed users:

    def notify_users(self, thread, comment):
   ...
                    body = self.format(thread, comment, parent_comment, email, admin=False)
   ...

It assigns a valid value to `parent_comment`, and also a valid value to `recipient`.  

If in the future it calls the `format` function with a valid value assigned to `parent_comment` when sending mails to the admin, I could change `admin = not parent_comment` to `admin = not recipient`, 'cause `recipient` is always the value only used for user notification. Anyway `admin` parameter is really niche.  

To-do:  
[ ] Eliminate `self.no_name` for now.

jelmer:
Please reply to comments in the GitHub system, rather than as one big comment - that makes the conversation harder.

You shouldn't need to add test-specific code, and certainly not based on paths. If you do need test-specific overrides, use e.g. mocks or monkey patching. 

Re the behaviour related to parent_comment: You're adding magic behaviour based on some unrelated variable; that makes the code harder to understand, and it's not necessary for the changes in this pull request either.

SnorlaxYum:
Okay, added admin parameter back and removed the test-specific code.

docs/docs/configuration/server.rst
```diff
@@ -229,6 +229,50 @@ timeout
...
+    {replier}
+        The name of the author of the comment. If the author has no name,
+        it will fallback to the term "Anonymous".
+
+    {repliee}
+        The author whose comment is replied to.  If the author has no
+        name, it will fallback to the term "Anonymous".
+
+    {receiver}
+        The receiver of the email.
```
jelmer: Their name, email?

isso/ext/notifications.py
```diff
@@ -139,17 +139,43 @@ def format(self, thread, comment, parent_comment, recipient=None, admin=False):
 
         else:
             uri = self.public_endpoint + "/id/%i" % parent_comment["id"]
-            key = self.isso.sign(('unsubscribe', recipient))
+            key = self.isso.sign(('unsubscribe', recipient["email"]))
```
jelmer: Why is this now looking at ["email"] ? Was the code broken previously?

isso/ext/notifications.py
```diff
@@ -139,17 +139,43 @@ def format(self, thread, comment, parent_comment, recipient=None, admin=False):
...
+                title=thread["title"],
+                replier=comment["author"] or "Anonymous")
+        else:
+            subject_format = list(self.isso.conf.getiter("mail", "subject_user"))
+            if len(subject_format) == 1:
+                return subject_format[0].format(
+                    title=thread["title"],
+                    repliee=parent_comment["author"] or "Anonymous",
+                    replier=comment["author"] or "Anonymous",
+                    receiver=recipient["author"] or "Anonymous")
```
jelmer: How can receiver be anonymous?

isso/ext/notifications.py
```diff
@@ -139,17 +139,43 @@ def format(self, thread, comment, parent_comment, recipient=None, admin=False):
...
         return rv.read()
 
+    def notify_subject(self, thread, comment, parent_comment=None, recipient=None, admin=False):
+        if admin:
+            return self.isso.conf.get("mail", "subject_admin").format(
+                title=thread["title"],
+                replier=comment["author"] or "Anonymous")
+        else:
+            subject_format = list(self.isso.conf.getiter("mail", "subject_user"))
+            if len(subject_format) == 1:
```
jelmer: When would subject_format not have length 1?

isso/ext/notifications.py
```diff
@@ -168,8 +194,8 @@ def notify_users(self, thread, comment):
                 email = comment_to_notify["email"]
                 if "email" in comment_to_notify and comment_to_notify["notification"] and email not in notified \
                         and comment_to_notify["id"] != comment["id"] and email != comment["email"]:
-                    body = self.format(thread, comment, parent_comment, email, admin=False)
-                    subject = "Re: New comment posted on %s" % thread["title"]
+                    body = self.format(thread, comment, parent_comment, comment_to_notify)
```
jelmer: comment_to_notify is passed in to the "recipient" argument. That seems wrong?

isso/ext/notifications.py
```diff
@@ -168,8 +194,8 @@ def notify_users(self, thread, comment):
...
+                    subject = self.notify_subject(thread, comment, parent_comment, comment_to_notify)
```
jelmer: Same here

docs/docs/configuration/server.rst
SnorlaxYum: Or you want me to s/The receiver/The name of the recipient ?

isso/ext/notifications.py
SnorlaxYum: Because I pass the whole comment record of the recipient instead of the email address to it, as you can see below.

isso/ext/notifications.py
SnorlaxYum: I use getiter instead of get to get a list of formats. When recipient and parent_comment is the same comment and subject_format has two or more items, then use the second one in the list, otherwise use the first. If subject_format has one item, then the only item will be used in all cases. You might wonder when recipient and parent_comment is different, it's what happens when a subcriber of a reply to the  parent_comment receives the mail about a new reply to the parent_comment.

isso/ext/notifications.py
SnorlaxYum: That's in a title. A receiver can be named "Anonymous" without a name.

isso/ext/notifications.py
SnorlaxYum: That's not wrong, the whole comment record of the recipient is passed to the parameter recipient. Then things like recipient["id"] in the record can be used, not just recipient["email"]

isso/ext/notifications.py
SnorlaxYum: Similar as above.

SnorlaxYum:
Replied to your comment above, about the parameter recipient that's not a mistake. I've tested and used in the branch of the original PR. And I'm running the branch of the original PR on my server. It works well.

docs/docs/configuration/server.rst
jelmer: If that's what the variable gets set to, yes.

isso/ext/notifications.py
jelmer: So the recipient argument is now misnamed?

docs/docs/configuration/server.rst
SnorlaxYum: Okay, I'll change it.

isso/ext/notifications.py
SnorlaxYum: So maybe change it to `recipient_comment`?

isso/ext/notifications.py
```diff
@@ -139,17 +139,42 @@ def format(self, thread, comment, parent_comment, recipient=None, admin=False):
...
 
-            rv.write("Unsubscribe from this conversation: %s\n" % (uri + "/unsubscribe/" + quote(recipient) + "/" + key))
+            rv.write("Unsubscribe from this conversation: %s\n" % (uri + "/unsubscribe/" + quote(recipient_comment["email"]) + "/" + key))
 
         rv.seek(0)
         return rv.read()
 
+    def notify_subject(self, thread, comment, parent_comment=None, recipient_comment=None, admin=False):
+        if admin:
+            return self.isso.conf.get("mail", "subject_admin").format(
```
jelmer: What happens if subject_admin is not set?

docs/docs/configuration/server.rst
```diff
@@ -229,6 +229,50 @@ timeout
...
+    {title}
+        The title of the isso thread.
+
+    {replier}
+        The name of the author of the comment. If the author has no name,
+        it will fallback to the term "Anonymous".
+
+subject_user
+    specify the subject format of the notification email sent to the subscribed
+    commenter. If it is set to two values, then the first one will be used when
```
jelmer: Why not just have two settings, one for new messages and one for replies?

isso/ext/notifications.py
```diff
@@ -139,17 +139,42 @@ def format(self, thread, comment, parent_comment, recipient=None, admin=False):
...
+                title=thread["title"],
+                replier=comment["author"] or "Anonymous")
+        subject_format = list(self.isso.conf.getiter("mail", "subject_user"))
```
jelmer: What if subject_user is not set? What if there are more than 2 items in the list?

isso/ext/notifications.py
```diff
@@ -139,17 +139,42 @@ def format(self, thread, comment, parent_comment, recipient=None, admin=False):
...
+            rv.write("Unsubscribe from this conversation: %s\n" % (uri + "/unsubscribe/" + quote(recipient_comment["email"]) + "/" + key))
 
         rv.seek(0)
         return rv.read()
 
+    def notify_subject(self, thread, comment, parent_comment=None, recipient_comment=None, admin=False):
+        if admin:
+            return self.isso.conf.get("mail", "subject_admin").format(
+                title=thread["title"],
+                replier=comment["author"] or "Anonymous")
```
jelmer: As mentioned earlier, perhaps use a constant for "Anonymous" (not a member variable, as you had earlier)

isso/ext/notifications.py
SnorlaxYum: The default fallback settings of the options are set in `share/isso.conf` . Refer to that file to check what will happen in that case.

docs/docs/configuration/server.rst
SnorlaxYum: Okay, I'll seperate it into two options.

isso/ext/notifications.py
SnorlaxYum: Re the first question: Similar to the answer to `What happens if subject_admin is not set?` above.  
Re the second question: Only the first two items will be used.

isso/ext/notifications.py
SnorlaxYum: Okay, since there's no such thing called constant in python, I might implement it by a function returning a certain string.

SnorlaxYum: For translation of `Anonymous` in the future I might still use a member variable. 'Cause in that case every time I call the function is a query from i18n inside isso, so in that case I think it will be better to assign a string to a member variable according to the language.

aquatix:
Any news on this? This would be a great addition.

SnorlaxYum:
Hmm...... Haven't received a feedback since I submitted my last commit. Whether it merged into the `master` branch is up to the reviewers. I'd like to make requested changes to this.

zhangnew:
Any news on this?

ix5:
@SnorlaxYum I have pushed a rebased version with the fixes to things I'd have lamented in review to https://github.com/ix5/isso/tree/pull-553-rebased

If you're happy with that, I'm going to merge it after conferring with @jelmer one last time.

jelmer:
Sorry, I'll leave this to you to review @ix5 - I've run out of steam on this PR. It's in a much better shape, but I'm still skeptical about the usefulness of making all of this configurable - and there are consequences of landing this, like making it much harder to change the format of emails in the future or code paths that only get exercised when certain configuration options are set since the tests only cover some fraction of the code.

I'll abstain.

ix5:
Yeah, on further reflection, I have to agree with jelmer. The proper way is to figure out a plugin/extension system (e.g. pluggable email module) instead of convoluting the config options even further.

PR #559, opened 2019-07-12
[WIP] add simple systemd startup script with no dependencies

Konzertheld:
The docs were lacking a systemd script that just starts isso without gunicorn so I thought I could add mine. If it is not wanted to have the file in the main repo, please deny this PR with an according comment and I will instead add it to my fork and add a link to the docs.

### DIFF
--- a/share/isso.service
+++ b/share/isso.service
@@ -1,0 +1,10 @@
+[Unit]
+Description=lightweight Disqus alternative
+
+[Service]
+Type=simple
+User=isso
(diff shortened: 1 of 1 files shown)


jelmer:
This seems reasonable, but please add some comments - i.e. /opt/isso/bin should be where isso was installed, etc.

Konzertheld:
Comments added, if you want you can re-open this.

Issue #301, opened 2017-01-22
Remove 404 when no comments


claudio4:
When there are not comments the server response is a 404 error and this is bad beacause de browsers and the search engines detect it as a broken link, please replace the 404 error to something like *no comments*.

Tanks for your time and nice project

Tetsumaki:
+1 for this issue please.

zontarian:
Yes, an empty array would be preferable or something like that.

jelmer:
How would isso know what valid pages are? The alternative is returning a 200, but also for pages that don't actually exist.

claudio4:
A option could be return a 204 (No Content).

jelmer:
Still, how would isso know which URLs are for valid pages and which ones aren't? Presumably you'd only want to return a 204 for comments on pages that exist, but a 404 for pages that don't exist.

claudio4:
Which is the point of knwing if a page doesn't exists if you can send a comment anyway and create it?

jelmer:
That's a fair point.

ix5:
Fixed by https://github.com/posativ/isso/pull/565

PR #565, opened 2019-08-11
Return 200 with empty array when there are no comments #301

tdro:
For some of us who are picky about 404s, the API returns 404 when there are no comments. The not found exception was removed and the endpoint returns empty as expected with HTTP status 200.

`{"id": null, "total_replies": 0, "hidden_replies": 0, "replies": []}`


jelmer:
Please also update the tests.

PR #579, opened 2019-11-01
Telegram bot notifications

hrantzsch:
This PR outlines the basic functionality to inform users about new comments via a Telegram bot. Would you be interested in integrating this in Isso?

As it is drafted here, the code requires the `python-telegram-bot` package. I would make the dependency optional and document it before this could be merged.


jelmer:
I think I would rather see some standard interface for notifications be defined that e.g. telegram integration could use.

PR #581, opened 2019-11-09
Mention official Debian repo instead of 3rd party

JocelynDelalande:
- isso is now packaged in official debian
- packages.crapouillou.net is no longer maintained

what else ? :)


jelmer:
Thanks!

PR #586, opened 2019-11-22
Auto moderate known authors

cpbotha:
This implements an option for the wordpress moderation feature where new comments by authors who have had approved comments are auto-approved.

In addition, I have made it so that when this option is active, the author needs to have had an approved comment within the last 6 months.

I have been testing this on https://cpbotha.net/ for almost two weeks now.

### DIFF
--- a/isso/db/comments.py
+++ b/isso/db/comments.py
@@ -86,9 +86,10 @@ class Comments:
         Search for previously activated comments with this author email.
         """
 
-        # if the user has not entered email, email == None in which case we can't check if they have previous comments
+        # if the user has not entered email, email == None, in which case we can't check if they have previous comments
         # in addition, we will only check for a matching email if the passed string is 3 chars or longer.
-        # we can only do len() on a string, so we use that check, which of course also ensures it's not None
+        # here we ensure that we only do this check if email is a string, which both excludes None,
+        # and ensures that this code correctly ignores other types that might accidentally be passed
         if type(email) is str and len(email) >= 3:
             # search for any activated comments within the last 6 months by email
             # this SQL should be one of the fastest ways of doing this check
--- a/isso/db/comments.py
+++ b/isso/db/comments.py
@@ -86,6 +86,9 @@ class Comments:
         Search for previously activated comments with this author email.
         """
 
+        # if the user has not entered email, email == None in which case we can't check if they have previous comments
+        # in addition, we will only check for a matching email if the passed string is 3 chars or longer.
+        # we can only do len() on a string, so we use that check, which of course also ensures it's not None
         if type(email) is str and len(email) >= 3:
             # search for any activated comments within the last 6 months by email
             # this SQL should be one of the fastest ways of doing this check
--- a/isso/db/comments.py
+++ b/isso/db/comments.py
@@ -87,12 +87,14 @@ class Comments:
         """
 
         if type(email) is str and len(email) >= 3:
-            # this SQL should be one of the fastest ways of determining if
-            # there's at least one comment with the given email and activation
+            # search for any activated comments within the last 6 months by email
+            # this SQL should be one of the fastest ways of doing this check
             # https://stackoverflow.com/questions/18114458/fastest-way-to-determine-if-record-exists
-            rv = self.db.execute(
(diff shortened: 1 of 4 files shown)


isso/db/comments.py
```diff
@@ -81,6 +81,25 @@ def activate(self, id):
             '    mode=1',
             'WHERE id=? AND mode=2'], (id, ))
 
+    def is_previously_approved_author(self, email):
+        """
+        Search for previously activated comments with this author email.
+        """
+
+        if type(email) is str and len(email) >= 3:
```
jelmer: What is this line trying to guard against?

share/isso.conf
```diff
@@ -71,6 +71,10 @@ password = please_choose_a_strong_password
 # them.
 enabled = false
 
+# with moderation enabled, automatically approve new posts by a
+# previously approved author
+activate-if-email-previously-approved = false
```
jelmer: Should this be called approve rather than activate? The term activate is a bit confusing to me (but I'm not a native English speaker).

jelmer: Please also add documentation for this new option.

cpbotha: Hey there @jelmer I agree with you on approve vs activate. I chose activate, because that's the wording used in the moderation emails. It sounds like you prefer "approve" so I'll change it, OK?

(P.S. we hadden dit ook in het Nederlands kunnen bespreken, maar dat is niet zo leuk voor de rest. ;)

isso/db/comments.py
cpbotha: in some setups, email is not required, in which case this function gets a None (see views/comments.py), hence the `str` check.

The `len(email) >= 3` is to guard against cases where an invalid email, or even a blank string was somehow passed.

share/isso.conf
cpbotha: I will add some documentation to the `server.rst` and push within about 10 minutes.

isso/db/comments.py
jelmer: Can we just explicitly check for None and perhaps add a comment about the length? That will hopefully easier to understand for others who read this code later.

cpbotha:
Hi there @jelmer. I have added a comment to the `author` field sanity check explaining the motivation.

I would prefer keeping the `str` type check together with the `len`, as this is the only watertight method to check for both `!= None`, *and*, following the tenets of defensive programming, to ensure that if some other type ever got passed in the code would handle that without raising an exception.

cpbotha:
I refined the wording of the comment: https://github.com/posativ/isso/pull/586/commits/33f35da3f49558f8c7a75af11c05f9c78b7e4b48

jelmer:
Thanks, the pull request looks good. How do you make sure that the e-mail address that somebody enters actually belongs to them (like on Wordpress)? Isn't it possible for a comment poster to just enter an email address of somebody they know had a comment approved previously (like the site owner) ?

cpbotha:
This is how the self-hosted wordpress also does it when you configure that setting. If someone could correctly guess the email that one of your commenters has used, the comment will be auto-approved.

In practice, finding the correct email addresses of pre-approved commenters is not straight-forward, and the spammers are not yet putting this level of effort in.

For smaller blogs this strikes a really good balance between usability and security. Before I switched to hugo and isso, this is the setting I've been using for 10 years+ on 3 of my most active wordpress sites, with no issues.

jelmer:
Ah, fair enough.

share/isso.conf
```diff
@@ -71,6 +71,10 @@ password = please_choose_a_strong_password
 # them.
 enabled = false
 
+# with moderation enabled, automatically approve new comments by an
+# author if they've had comments approved within the last 6 months
```
jelmer: Can you add a comment here about nothing verifying the e-mail address?

jelmer:
FWIW I'm happy to merge this once the two remaining comments are addressed.

cpbotha:
Hi there @jelmer -- I have added an explanation of the unverified email address in isso.conf and also in the server.rst documentation.

jelmer:
Cherry-picked, thanks!

Issue #584, opened 2019-11-20
Support Content Security Policy (CSP) without `style-src 'unsafe-inline'`


jwatt:
The way Isso currently works is to insert a `<style>` element with inline styles into the document. That means that Isso can't be used on a site that uses CSP without it using `style-src 'unsafe-inline'` ([docs](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/style-src)).

Since Isso serves the file `js/embed.min.js`, presumably it should be possible to fix this issue by having it also serve the style rules as a separate style sheet instead of as inline style. (Hopefully, at least.)

PR #597, opened 2019-11-28
Make Isso compatible with Content Security Policies without `script-src 'unsafe-inline'`

jwatt:
This allow Isso to be used with websites that have a Content Security Policy that doesn't include `script-src 'unsafe-inline'`.


jwatt:
(Note that this is not #584, which is about style, not script. I'll do style after this.)

Issue #570, opened 2019-09-02
Admin comment edits strips new lines


skylord123:
When I edit comments from the admin UI it removes all newlines that were in the comment before I modified it. It's strange because you will save it and it will show the newlines but once you refresh it will show the same comment with them removed. 

Admin editor should match the same functionality as the client side one.

As a work-around you have to convert all newlines to `<br />` and it will work then. This is real annoying.

ix5:
As a crutch, you could wrap the area with `<pre>` tags and strip them before submitting.
But bringing the client-side editor into the admin area would be nicer.

ix5:
Do give https://github.com/posativ/isso/pull/604 a try.

hamoid:
Here an example of how a comment looks in admin. All line breaks are gone in the admin, but not in the comment shown publicly. If I edit and save in the admin, the one-lineness is saved :(

![2020-04-23-123246_1120x201_scrot](https://user-images.githubusercontent.com/108264/80089817-c8c51280-855e-11ea-878d-7d3fa5df58bf.png)

So there are two aspects: one is how comments are shown in the admin list, and the other is how it is shown in the admin editor (and later saved).

ix5:
@skylord123 fixed by #604, would you mind closing this?

PR #604, opened 2019-12-28
Use <pre> for comments in admin area

ix5:
Also update `admin.css` so that lines get wrapped

Note that `document.getElementById().textContent()` already overlooks the `<pre>` tags, so there is no need to strip them when using [send_edit()](https://github.com/posativ/isso/blob/5655f194b311c50126713e6f40440e7f89be6479/isso/js/admin.js#L92).

This is a dirty fix, but adding the client-side `contenteditable` mechanism here is something I'm not prepared to do. If anyone wants to pick this up, feel free to do so!

### DIFF
--- a/isso/css/admin.css
+++ b/isso/css/admin.css
@@ -113,6 +113,9 @@ a {
   text-align: center;
   width: 100%;
 }
+.isso-comment pre {
+    white-space: pre-wrap;
+}
 .isso-comment-footer a {
   cursor: pointer;
 }
--- a/isso/templates/admin.html
+++ b/isso/templates/admin.html
@@ -123,7 +123,7 @@
         {% if comment.mode == 4 %}
           <strong>HIDDEN</strong>. Original text: <br />
         {% endif %}
-        <div id="isso-text-{{comment.id}}">{{comment.text}}</div>
+        <div id="isso-text-{{comment.id}}"><pre>{{comment.text}}</pre></div>
       </div>
       <div class='isso-comment-footer'>
         {% if conf.votes and comment.likes - comment.dislikes != 0 %}


ix5:
Hmm, since it seems no one has stepped up so far and provided a proper solution, I guess this is a good-enough stopgap one.

See https://github.com/posativ/isso/issues/570

hamoid:
Hi! Does this save the `<pre>` with the comment? And if you edit it multiple times, adds it each time?

Wouldn't it be better to do `<div id="isso-text-{{comment.id}}" class="pre">` and create a `.pre` css style? (I may be totally wrong)

ix5:
> Hi! Does this save the `<pre>` with the comment? And if you edit it multiple times, adds it each time?

We have a saying in German: "He who is able to read has the clear advantage".
See in OP:

> Note that `document.getElementById().textContent()` already overlooks the `<pre>` tags, so there is no need to strip them when using `send_edit()`.

So rest assured, I already tested that "case" and am running it locally - it's working just fine.

> Wouldn't it be better to do `<div id="isso-text-{{comment.id}}" class="pre">` and create a `.pre` css style? 

Well you _want_ the browser to explicitly render _and treat internally_ the text field as a `<pre>` element so that with the `contenteditable` property, you can edit the contents directly while preserving the line breaks etc.
If you only style it with `pre` styles, your nice manualy formatting, including the line breaks, will get stripped out, because in HTML, any amount of whitespace is collapsed into one.

hamoid:
And he who has time to read has even more advantage :-P

ix5:
@jelmer thoughts?

jelmer:
Why <pre> rather than proper rendering?

ix5:
> Why `<pre>` rather than proper rendering?

Because then you'd have to deal with switching back and forth between editing markdown source and rendered HTML inside the admin interface - or worse, implement some variation of `contenteditable` with something like CKEditor or [QuillJS](https://quilljs.com/).

IMO the admin interface should be kept simple, and admins can be expected to edit markdown themselves.

ix5:
@fluffy-critter since you're taking a look at issues anyway, what are your thoughts? This is way easier than implementing the render functionality of the client embed js in admin once again.

fluffy-critter:
I don't have any particular thoughts about this aspect.

jelmer:
Okay, let's land this for the moment in lieu of a proper fix

jelmer (approved):

ix5:
In conjuntion with 2391e8e7a90a1c9d859910ff7b343f33ae6999fb, this PR has opened up another annoying bug when editing comments from inside the admin area, appending literal `<pre>` tags when editing.

The admin really shouldn't be using `contenteditable` but rather proper forms, else we're opening a whole can of worms.

PR #605, opened 2019-12-28
html: Add markdown render flags option

ix5:
New `flags` option in the `[markdown]` section to customize Misaka's Markdown HTML rendering.

By default, no flags are set.

```
[markup]
flags = skip-html, escape, hard-wrap
```

Update `CHANGES.rst` and add the section to the sample `isso.conf` and `isso-dev.conf`.

---

Also fix flaky tests and update to python 3.8.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,6 +12,9 @@ matrix:
     - python: 3.7
       dist: xenial
       env: TOX_ENV=py37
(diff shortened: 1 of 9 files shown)


jelmer:
Thanks! Merged in #616

ix5:
Great, thank you!

PR #610, opened 2020-01-24
New 'latest' endpoint to serve latest comments. Fixes #556.

### LINKED ISSUE #556: List the latest comments (no matter the URI)
I have a blog and I'm using Isso (thanks for this software, is great!!).

I want to implement a section in my blog that should show the latest 10 comments I got in the whole blog, with a link to the original post).

For that, I'd need to list all the comments, not filtering by a specific URI. Of course, not *all* the comments, the latest N.

Is there a current way of doing this?

If not, I'd propose to extend the API, add a new endpoint "/latest/N" where the N is limit (it's mandatory! it doesn't make sense to get *everything*). 

I would work on adding this new endpoint, but I'd love to get some feedback about this proposal, ideas, etc.

Thanks!!

facundobatista:


isso/tests/test_comments.py
```diff
@@ -33,6 +33,7 @@ def setUp(self):
         conf.set("general", "dbpath", self.path)
         conf.set("guard", "enabled", "off")
         conf.set("hash", "algorithm", "none")
+        conf.set("general", "latest-enabled", "true")
```
jelmer: Please also add a short entry to docs/docs/configuration/server.rst about this new configuration option.

PR #632, opened 2020-04-08
Feature : filtering comments based on URI of their pages

babakRz63ai:
Site managers have an option to set a filter for acceptable URI's. You need to put a Regex pattern as new field `uri-filter` under section `[guard]`

### DIFF
--- a/About-URI-filtering.md
+++ b/About-URI-filtering.md
@@ -1,0 +1,3 @@
+# About URI filtering
+
+This feature causes *Isso* to accept or reject comments based on URI of a page a new comment is supposed to belong to. This is done by mentioning a pattern in `[guard]` section of the server config file. 
--- a/About-URI-filtering.md
+++ b/About-URI-filtering.md
@@ -1,3 +1,36 @@
 # About URI filtering
 
-This feature causes *Isso* to accept or reject comments based on URI of a page a new comment is supposed to belong to. This is done by mentioning a pattern in `[guard]` section of the server config file. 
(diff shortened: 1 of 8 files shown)


ix5:
Seems like what you're trying to accomplish could be better implemented outside isso. Your PR adds a lot of complexity.

Have you thought about *not* including the isso `embed.js` for the pages where you want it to not display/allow comments?

babakRz63ai:
> Have you thought about not including the isso embed.js for the pages where you want it to not display/allow comments?

Of course I did, but it is limited to the client side only. I considered possibility to post comments independent of website pages and using a tool like `curl` for posts I don't want any comment at, can be somewhat inconvenient for me as a site admin.

ix5:
So you think someone is going to

- Reverse-engineer your isso embed to figure out your isso instance URL
- Craft a specific curl/whatever request to post a comment to somewhere you don't want any comments
- Profit?

That sounds very unlikely. And again, you can "protect" yourself by not including the isso embed.js where you don't want anyone commenting. Because even if an "attacker" manages to post a comment using curl, no one will see it anyway.

Or am I misunderstanding your intention, you mean that you want to restrict certain pages to not *allow* comments, but still *display* them? So that only you as a site admin can post comments? For that, why not use the moderation feature? You can disallow any comments you don't want then.

In any case, you PR adds complexity where other solutions offer a better way.

jelmer:
I'll close this; @ix5 makes some good points, and there hasn't been any follow-up in the last year.

PR #636, opened 2020-04-15
Using .access_route instead of .remote_addr to take into account HTTP_X_FORWARDED_FOR header

Lucas-C:


isso/views/comments.py
```diff
@@ -275,7 +275,7 @@ def new(self, environ, request, uri):
             data["website"] = normalize(data["website"])
 
         data['mode'] = 2 if self.moderated else 1
-        data['remote_addr'] = utils.anonymize(str(request.remote_addr))
+        data['remote_addr'] = utils.anonymize(str(request.access_route[0]))  # Using .access_route instead of .remote_addr to take into account HTTP_X_FORWARDED_FOR header
```
jelmer, line 278: Wouldn't this allow users to vote multiple times by spoofing the X-Forwarded-For header?

cpbotha:
I run isso behind uwsgi and nginx, so ALL of my notification emails have `127.0.0.1` as the IP address, which adds no information whatsoever.

I guess many of the isso deployments employ some sort of reverse proxy configuration, so more people will have this problem.

With regard to spoofing: How often are people going to try that on real-world isso installations?! Furthermore, if they try to spoof the header, won't the reverse proxy overwrite it with correct information in any case?

I  would be super grateful if you could merge this PR.

cpbotha:
Here's a solid-looking SO answer on how to access `request.access_route` securely: https://stackoverflow.com/a/22936947/532513

``` python
trusted_proxies = {'127.0.0.1'}  # define your own set
route = request.access_route + [request.remote_addr]

remote_addr = next((addr for addr in reversed(route) 
                    if addr not in trusted_proxies), request.remote_addr)
```

Lucas-C:
> won't the reverse proxy overwrite it with correct information in any case?

I agree. I am no nginx expert, but I think the hypothetical spoofing issue should be handled on the nginx configuration side, not by isso : https://serverfault.com/a/414166/97081

Lucas-C:
> Here's a solid-looking SO answer on how to access request.access_route securely

Then, with this solution, a new isso variable should be introduced to set the list of trusted proxies ?

Lucas-C:
The HTTP `X-Forward-*` headers, and the spoofing scenarios, are quite a complex subject IMHO.
E.g. I just learned about a new standard header from the nginx docs : https://www.nginx.com/resources/wiki/start/topics/examples/forwarded/ (RFC 7239)

Anyway, I'd vote for adding the "secure" solution from the SO answer @cpbotha mentioned, but only enable it if a given "trusted proxies" isso configuration variable is set. This way it will remain fully backward compatible, and "forces" the isso admin to define a list of trusted proxies in order to enable this behaviour.

cpbotha:
> > Here's a solid-looking SO answer on how to access request.access_route securely
> 
> Then, with this solution, a new isso variable should be introduced to set the list of trusted proxies ?

To me that indeed sounds like the best solution. Could it be that having by default 127.0.0.1 in that set would fix the issue for a large part of cases?

isso/views/comments.py
Lucas-C: Replies were made in the PR discussion

cpbotha:
> Anyway, I'd vote for adding the "secure" solution from the SO answer @cpbotha mentioned, but only enable it if a given "trusted proxies" isso configuration variable is set. This way it will remain fully backward compatible, and "forces" the isso admin to define a list of trusted proxies in order to enable this behaviour.

Wow that sounds perfect to me! I'm just an innocent bystander and occasional code contributor here. :)

@jelmer what do you think?

Lucas-C:
Not in mine ^^

I'd still prefer this to be a "manual opt-in" in order to limit spoofing risks and remain fully backward compatible. However this should be clearly documented, maybe in a new "Proxies" section of https://posativ.org/isso/docs/configuration/server/ ?

jelmer:
> Anyway, I'd vote for adding the "secure" solution from the SO answer @cpbotha mentioned, but only enable it if a given "trusted proxies" isso configuration variable is set. This way it will remain fully backward compatible, and "forces" the isso admin to define a list of trusted proxies in order to enable this behaviour.

Yep, that seems like the correct way to do this to me as well.

Lucas-C:
I have updated this PR accordingly

isso/views/comments.py
```diff
@@ -336,6 +337,20 @@ def new(self, environ, request, uri):
         resp.headers.add("X-Set-Cookie", cookie("isso-%i" % rv["id"]))
         return resp
 
+    def _remote_addr(self, request):
+        '''
```
jelmer, line 341: Please use regular docstring conventions (https://www.python.org/dev/peps/pep-0257/), i.e. """ rather than ''' and a first line standalone summary.

share/isso.conf
```diff
@@ -112,6 +112,13 @@ reload = off
 # in production.
 profile = off
 
+# an optional list of reverse proxies IPs behind which you have deployed
+# your Isso web service (e.g. `127.0.0.1`).
+# This allow for proper remote address resolution based on a
+# `X-Forwarded-For` HTTP header, which is important for the mechanism
+# forbiding several comment votes coming from the same IP.
```
jelmer, line 119: s/forbiding/forbidding/

I think this should say subnet rather than IP since it calls utils.anonimize().

CONTRIBUTORS.txt
```diff
@@ -74,6 +74,10 @@ In chronological order:
 * Lucas Cimon @Lucas-C
   * Added the possibility to define CORS origins through ISSO_CORS_ORIGIN environment variable
   * Fix a bug with <a> in <svg>
+  * Fixing likes counter of replies not being displayed
```
jelmer, line 77: Nit: please avoid gerunds, just use active verbs - i.e. "Fix ..." rather than "Fixing ..."

share/isso.conf
Lucas-C: 👍

isso/views/comments.py
Lucas-C: Ok, improved !

Lucas-C:
PR updated

Lucas-C:
Thanks for reviewing & merging this !

PR #646, opened 2020-04-29
Ok is not OK. Fixes #347

### LINKED ISSUE #347: Isso Forces CORS Preflight but is unable to handle CORS Preflight requests

Konzertheld:


Konzertheld:
In Python 2 on isso 0.12.2, that is.

jelmer:
I've only briefly glanced at the issue; how does this fix the issue? Which code expects "OK" rather than "Ok" ?

Konzertheld:
I just came across that issue (and the related ones) because I had CORS issues myself, seeing the preflight checks fail as their response was not "200 OK" - that's what I saw in the browser console. So I dug around in the code a little bit at the places discussed in the ticket and noticed the case was not correct. Changed it to OK and my problems went away. I never investigated the rest of the CORS handling.

jelmer:
I'm hesitant about merging this without understanding how it fixes things - it seems wrong for other software to rely on specific capitalization.

Konzertheld:
It's not completely random to use the capitalization - it's how it is defined in the HTTP protocol: https://tools.ietf.org/html/rfc2616#section-10.2.1

jelmer:
On Sun, Oct 11, 2020 at 05:17:34PM -0700, Christian wrote:
> It's not completely random to use the capitalization - it's how it is defined in the HTTP protocol: https://tools.ietf.org/html/rfc2616#section-10.2.1
Yes, but the RFC also explicitly says that the reason phrase is for
human consumption and can be changed:

https://tools.ietf.org/html/rfc2616#section-6.1.1

"The reason phrases listed here are only recommendations -- they MAY
be replaced by local equivalents without affecting the protocol."

So whatever is expecting to see "OK" here with that specific
capitalization is violating the spec.

-- 
Jelmer Vernooĳ <jelmer@jelmer.uk>
PGP Key: https://www.jelmer.uk/D729A457.asc

blatinier:
I agree on this with @jelmer It's not meant to be used by computers.
However I can guess how a poorly designed system could rely on this and it doesn't hurt to merge it anyway…

PR #649, opened 2020-05-09
admin: Set default order_by to 'created' (desc)

ix5:
This shows recent comments first in the admin panel.

Personal preference of mine, but maybe others agree?


jelmer:
Yeah, this is my preference as well.

ix5:
Neat, thanks :)

Issue #521, opened 2019-01-03
Browsers without language set won't display Isso comments


0-1:
If a browser has no default language set, Isso won't work.

TypeError: undefined has no properties.
embed.min.js:formatted:245

Points to this line:
`lang: (navigator.language || navigator.userLanguage).split('-') [0],`

You might want to set a default language for isso if you need it.

ix5:
Closing since https://github.com/posativ/isso/pull/674 was merged recently.
@0-1 feel free to re-open in case this is still an issue.

PR #674, opened 2020-11-04
Handle browsers with no navigator.language better (#521)

zackw:
The code for setting `config["lang"]`, in `isso/app/config.js`,
assumes that browsers will always provide a value for
`navigator.language` and/or `navigator.userLanguage`.  Per bug #521,
this is not a safe assumption.

While I was attempting to fix this, I noticed that regional variants
of a language (`zh-TW`, `pt-BR`) were being handled in an ad-hoc,
unreliable manner.  I also noticed a new user-agent language property
[`navigator.languages`][] which more closely matches the semantics of
[`Accept-Language`][]—it would be good to support that.

This patch addresses all the above problems, as follows:

1. Add a new configuration property `data-isso-default-lang` that
   specifies the language to use (instead of English) when the browser
   *doesn’t* have a preference.

2. Document that we expect the value of `data-isso-lang` and
   `data-isso-default-lang` to be a [BCP 47][] language tag, because
   this is what `navigator.language` etc contain.  (The practical
   upshot is that tags like `zh-TW` are officially allowed now.)

3. In `config.js`, compile an array of candidate language tags, in
   descending order of preference, from all available sources:
   `data-isso-lang`, `navigator.languages`, `navigator.language`,
   `navigator.userLanguage`, `data-isso-default-lang`, and finally
   `"en"` as a backstop.  Handle any or all of the above being
   null/undefined/empty.  This array goes into `config["langs"]`.
   `config["lang"]` is removed.

4. In `i18n.js`, select the first entry in `config["langs"]` for which
   we have both pluralforms and a translation, and make that the value
   of `i18n.lang`.  An English backstop is supplied here too for extra
   defensiveness.  Also, if we don’t have a translation for say
   `zh-HK`, try `zh` before moving on to the next entry in the array.

5. New function `utils.normalize_bcp47` ensures that we process
   language tags, whereever they came from, case-insensitively and
   treating `_` as equivalent to `-`.

6. Move `utils.ago` to `i18n.ago` to avoid a circular dependency
   between utils and i18n.

[`navigator.languages`]: https://developer.mozilla.org/en-US/docs/Web/API/NavigatorLanguage/languages
[`Accept-Language`]: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Accept-Language
[BCP 47]: https://tools.ietf.org/html/bcp47

### DIFF
--- a/docs/docs/configuration/client.rst
+++ b/docs/docs/configuration/client.rst
@@ -121,12 +121,32 @@ Defaults to `true`.
 data-isso-lang
 --------------
 
-Override useragent's preferred language. Isso has been translated in over 12
-languages. The language is configured by its `ISO 639-1
-<https://en.wikipedia.org/wiki/ISO_639-1>`_ (two letter) code.
+Always render the Isso UI in this language, ignoring what the
+user-agent says is the preferred language.  The default is to
+honor the user-agent's preferred language, and this can be
+specified explicitly by using ``data-isso-lang=""``.
 
-You find a list of all supported languages on `GitHub
-<https://github.com/posativ/isso/tree/master/isso/js/app/i18n>`_.
+The value of this property should be a `BCP 47 language tag
+<https://tools.ietf.org/html/bcp47>`_, such as "en", "ru", or "pt-BR".
+Language tags are processed case-insensitively, and may use
+underscores as separators instead of dashes (e.g. "pt_br" is treated
+the same as same as "pt-BR").
+
+You can find a list of all supported languages by browsing the
+`i18n directory
+<https://github.com/posativ/isso/tree/master/isso/js/app/i18n>`_ of
+the source tree.
+
+data-isso-default-lang
+----------------------
+
+Render the Isso UI in this language when the user-agent does not
+specify a preferred language, or if the language it specifies is not
+supported.  Like ``data-isso-lang``, the value of this property should
+be a BCP 47 language tag.  Defaults to "en".
+
+If you specify both ``data-isso-default-lang`` and ``data-isso-lang``,
+``data-isso-lang`` takes precedence.
 
 data-isso-reply-to-self
 -----------------------
--- a/isso/js/app/i18n.js
+++ b/isso/js/app/i18n.js
@@ -8,7 +8,9 @@ define(["app/config", "app/i18n/bg", "app/i18n/cs", "app/i18n/da",
     "use strict";
 
     var pluralforms = function(lang) {
-        switch (lang) {
+        // we currently only need to look at the primary language
+        // subtag.
+        switch (lang.split("-", 1)[0]) {
         case "bg":
         case "cs":
         case "da":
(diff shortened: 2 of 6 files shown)


docs/docs/configuration/client.rst
```diff
@@ -65,12 +65,33 @@ Defaults to `true`.
...
+honor the user-agent's preferred language, and this can be
+specified explicitly by using ``data-isso-lang=""``.
+
+The value of this property should be a `BCP 47 language tag
+<https://tools.ietf.org/html/bcp47>`_, such as "en", "ru", or "pt-BR".
+Language tags are processed case-insensitively, and may use
+underscores as separators instead of dashes (e.g. "pt_br" is treated
+the same as same as "pt-BR").
+
+As of November 2020, Isso has been translated into 25 languages.
```
stefangehn, line 79: Since this line tends to get out of sync sooner or later maybe just omit it

isso/js/app/i18n.js
```diff
@@ -101,22 +93,46 @@ define(["app/config", "app/i18n/bg", "app/i18n/cs", "app/i18n/da",
...
     };
 
-    var plural = pluralforms(lang);
+    // for each entry in config.langs, see whether we have a catalogue
+    // entry and a pluralforms entry for it.  if we don't, try chopping
+    // off everything but the primary language subtag, before moving
+    // on to the next one.
+    var lang, plural, translations;
+    for (var i = 0; i < config.langs.length; i++) {
+        plural = pluralforms(lang);
```
ix5, line 113: This fails for me since `lang` is none.
```
Uncaught TypeError: lang is undefined
```
Did you mean to add `lang = config.langs[i];` before attempting to call `pluralforms(lang)`?

ix5:
Good job on the PR!

I've tested with `navigator.lang{ueages}` set to `en`, `en-US`, `de`, `de-de` and `pt-BR` so far.
Apart from the nit in the linline comment I posted, it works great.

Could you rebase on latest master please? The `pt_PT` translation was added in the meantime.
I've pushed a rebased version to [ix5/isso:navigator-lang-674](https://github.com/ix5/isso/tree/navigator-lang-674) for you to pick.

docs/docs/configuration/client.rst
zackw: Agreed.

isso/js/app/i18n.js
zackw: Doh, yes I did.  I do most of my programming in Python so I'm used to for-loops over arrays giving me the value, not the index.  Thanks for catching, will be fixed in the next push.

zackw:
@ix5 Rebased and pushed now.

ix5:
Great, thanks! @jelmer @blatinier please review when you find the time.

ix5:
@zackw I've rebased your branch on top of current master, please see https://github.com/ix5/isso/tree/pull-674-rebased
Would you mind adding an entry to the `CHANGES.rst` file as well?

After that, I'll merge this PR.

zackw:
@ix5 Done.

stefangehn:
LGTM, I think proper handling of the two-part language identifiers is also worth merging this.

ix5:
Thank you for reviewing, @stefangehn! 
And thank you for your patience @zackw!

PR #685, opened 2021-01-17
share: isso.conf: Use dashed-case for misaka 2.0

ix5:
Misaka expects extensions and flags in `dashed-case` instead of the legacy `snake_case` as of version 2.0.

See https://github.com/FSX/misaka/blob/v2.1.1/misaka/utils.py#L15-L37

Also align "options" values with the defaults in `utils/html.py`.

### DIFF
--- a/share/isso.conf
+++ b/share/isso.conf
@@ -188,9 +188,10 @@ require-email = false
 # Customize markup and sanitized HTML. Currently, only Markdown (via Misaka) is
 # supported, but new languages are relatively easy to add.
 
-# Misaka-specific Markdown extensions, all flags starting with EXT_ can be used
-# there, separated by comma.
-options = strikethrough, autolink, fenced_code, no_intra_emphasis
+# Misaka-specific Markdown extensions, all extensions can be used here,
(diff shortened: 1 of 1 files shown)


share/isso.conf
```diff
@@ -188,9 +188,10 @@ require-email = false
 # Customize markup and sanitized HTML. Currently, only Markdown (via Misaka) is
 # supported, but new languages are relatively easy to add.
 
-# Misaka-specific Markdown extensions, all flags starting with EXT_ can be used
-# there, separated by comma.
-options = strikethrough, autolink, fenced_code, no_intra_emphasis
+# Misaka-specific Markdown extensions, all extensions can be used here,
+# separated by comma, either by their name or by EXT_<extension>.
+# Careful: Misaka 1.0 used "snake_case", but 2.0 needs "dashed-case"!
```
jelmer, line 193: We've been on misaka 2.0 since 1.5, do we really need to mention misaka 1.0?

ix5: I was directing that comment towards people upgrading from older versions.

PR #695, opened 2021-02-01
Only use npm for managing JS frontend build dependencies

stefangehn:
Replace bower as well as global npm package installations in Dockerfile with a plain `package.json` that is used to configure what `npm install` will install. This avoids a dependency on bower which was deprecated in 2017 and hopefully makes it easier for others to build the JS frontend files.

There are lots of more things to cleanup in this area (jade got renamed to pug, requirejs could be replaced by webpack, just to name two outdated dependencies) so this is just a start.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -26,6 +26,8 @@
 /docs/_build
 /docs/_static/css/site.css
 
+node_modules/
+
 /pip-selfcheck.json
 
 # github/gitignore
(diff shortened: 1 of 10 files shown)


stefangehn:
I don't think this change fixes the FTBFS in Debian packaging already. That one looks very much like one of the required node packages is somehow broken or incompatible and I can neither reproduce it when installing dependencies via npm (i.e. this branch) or when redirecting everything to use globally installed dependencies (from Ubuntu 20.04 though so that may be closer to what Debian 10 ships).

Regardless of that I hope that avoiding bower is a first start in the right direction.

blatinier:
Yup great start!

PR #696, opened 2021-02-03
Add missing commata to setup.py classifiers

stefangehn:
Properly separate setup classifier strings by a comma, otherwise they
end up as a single string (can be observed in the output created by
`setup.py bdist_wheel`).


jelmer:
Sorry, we raced each other on this change - I added this when I noticed that it was broken as part of the 0.12.4 release.

stefangehn:
no problem as long as it's fixed :+1:

Issue #682, opened 2021-01-15
[Discussion][Multi-Site] Future of cookies with SameSite and Secure


ix5:
As of 2021-01, Firefox 84 warns when hosting isso on a subdomain and accessing cookies.

Repro:
- Isso hosted at `comments.example.com`
- "Blog" hosted at `example.com`

**Warning:**
```
Cookie “isso-[id]” will be soon rejected because it has the “SameSite” attribute set to “None”. [...]
```
with a link to [MDM SameSite cookies](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Set-Cookie/SameSite).

**Cause:**  
In [isso/views/comments.py](https://github.com/posativ/isso/blob/944e4846847abc9b252f9b016158e336ffb77443/isso/views/comments.py#L312-L333), e.g. `new()` creates a new cookie with `SameSite` implicitly set to `False` and `Secure` implicitly set to `None`[^1].

**Possible solutions**  
1. Isso sets `Secure` cookie by default. Big problem is that large parts of isso    assume that it might be accessed over `http`, see e.g. the    [default isso.conf](https://github.com/posativ/isso/blob/944e4846847abc9b252f9b016158e336ffb77443/share/isso.conf#L23-L27)
2. Isso conditionally sets `Secure` cookie only for `https` hosts. Ugly but    doable.

**Help wanted**  
Would be great if someone who's more involved with web technology could step in and give their ideas and clarify whether my understanding here is even correct.

[^1]: For the full call chain, see the later [X-Set-Cookie](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Set-Cookie) and the defaults of `werkzeug`'s [http.dump_cookie](https://werkzeug.palletsprojects.com/en/1.0.x/http/#werkzeug.http.dump_cookie)

Fryboyter:
> Big problem is that large parts of isso assume that it might be accessed over http, see e.g. the default isso.conf

According to the documentation, the entry for host with http only exists because of Firefox users who completely suppress the HTTP referer. As far as I know, this must be configured explicitly via about:config.

In my configuration, the entries were previously in reverse order. So first https and then http. There have been no complaints so far. I just tested it and completely removed the http entry from the configuration file. Isso continued to work with Firefox without any problems.

[...]

PR #639, opened 2020-04-16
Adding a [server] proxy-fix-enable-x-prefix configuration option

Lucas-C:
...to take into consideration the `X-Forwarded-Prefix` HTTP header

Issue #558, opened 2019-07-12
Issues with admin and https


Konzertheld:
In master, the admin login form on /admin sends the login credentials to http even when it is called on https. Also, css and svg files in admin won't load for me because of the mixed protocol. Apparently Firefox blocks loading unsecure resources on secured sites. I will investigate further and keep this ticket updated.

Konzertheld:
Workaround: Setting server.public-endpoint to https://comments.example.org without trailing slash

Konzertheld:
I figured that the issue is in `werkzeug.local` that is called in the mentioned line as a fallback for `public-endpoint`. The returned value for `local.host` has `http` as protocol even if the host defined in the config has `https` and the page is called via https.

Konzertheld:
Got it. `local.host` is set on dispatch, which uses the `host()` function in wsgi.py, precisely this line is responsible: https://github.com/posativ/isso/blob/faaff1d4acaad42fc6485ce2090875ade265ea43/isso/wsgi.py#L33

And the `url_scheme` referenced there comes down to http or https based on a variable `ssl_context` being set or not in the server class, which seems to be always unset, based on this line in the same file: https://github.com/posativ/isso/blob/faaff1d4acaad42fc6485ce2090875ade265ea43/isso/wsgi.py#L214

If I do SSL termination with Nginx reverse proxy, I use `X-Forwarded-Proto` which should be respected somewhere but is not.

Konzertheld:
To me this looks like the appropriate point where the X_FORWARDED_PROTO should be taken care of is in `werkzeug`. There are too many places where values set by werkzeug are used by isso to fix this in isso itself.

Konzertheld:
And indeed `werkzeug` does look at `X_FORWARDED_PROTO` but ignores it because it is set to trust 0 proxies. At which point I am out because I have no idea how to configure that in isso.

seyuf:
Thanks @Konzertheld for the workaround 👍 .
However, even after setting the public endpoint, the admin seems to be redirected from https to http after the login ...

Konzertheld:
@seyuf Did you include the https in your public-endpoint and host? I did, and I am not redirected.

ix5:
@Konzertheld this might give you an idea:

https://github.com/posativ/isso/blob/e6dbd21819f8fdca9fdace447e9ba01276e8d6a9/isso/__init__.py#L82-L87

[...]

PR #700, opened 2021-02-03
Improve cookie SameSite/secure handling

ix5:
Add new `samesite` option in `[server]` section to configure `SameSite` header for cookies.

As a fallback, use `local.host` to detect URL scheme and set `SameSite` to `None` (https) or `Lax` (http) accordingly.

Set `Secure` attribute in response header so that cookies will only be sent when requesting content from `https://`
URLs.

Fixes:
```
Cookie “isso-[id]” will be soon rejected because it has the “SameSite” attribute set to “None”
```

(werkzeug sets `SameSite` to `None` by default anyway)

**NOTE:** This change _might_ break quite a few local test setups!

Discussion: https://github.com/posativ/isso/issues/682

See: https://developer.mozilla.org/docs/Web/HTTP/Headers/Set-Cookie/SameSite
and https://werkzeug.palletsprojects.com/en/1.0.x/http/#werkzeug.http.dump_cookie

---

Sidenote:
- Setting `isso_host_script` should somehow be unified instead of duplicated in every function. `local.host` is only available per-request though, so there cannot be a class variable such as `self.isso_host_script`.

### DIFF
--- a/isso/views/comments.py
+++ b/isso/views/comments.py
@@ -309,10 +309,9 @@ class API(object):
         # notify extension, that the new comment has been successfully saved
         self.signal("comments.new:after-save", thread, rv)
 
-        cookie = functools.partial(dump_cookie,
-                                   value=self.isso.sign(
-                                       [rv["id"], sha1(rv["text"])]),
-                                   max_age=self.conf.getint('max-age'))
+        cookie = self.create_cookie(
+            value=self.isso.sign([rv["id"], sha1(rv["text"])]),
+            max_age=self.conf.getint('max-age'))
 
         rv["text"] = self.isso.render(rv["text"])
         rv["hash"] = self.hash(rv['email'] or rv['remote_addr'])
@@ -347,6 +346,24 @@ class API(object):
             remote_addr = next((addr for addr in reversed(route)
                                 if addr not in self.trusted_proxies), remote_addr)
         return utils.anonymize(str(remote_addr))
+
+    def create_cookie(self, **kwargs):
+        """
+        Setting cookies to SameSite=None requires "Secure" attribute.
+        For http-only, we need to override the dump_cookie() default SameSite=None
+        or the cookie will be rejected.
+        https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Set-Cookie/SameSite#samesitenone_requires_secure
+        """
+        isso_host_script = self.isso.conf.get("server", "public-endpoint") or local.host
+        samesite = self.isso.conf.get("server", "samesite")
+        if isso_host_script.startswith("https://"):
+            secure = True
+            samesite = samesite or "None"
+        else:
+            secure = False
+            samesite = samesite or "Lax"
+        return functools.partial(dump_cookie, **kwargs,
+                                 secure=secure, samesite=samesite)
 
     """
     @api {get} /id/:id view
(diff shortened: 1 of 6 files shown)


blatinier (approved):

ix5:
Will need to write tests for this.

ix5:
Tagging @Lucas-C (because of https://github.com/posativ/isso/pull/639), @Konzertheld (because of https://github.com/posativ/isso/issues/558) in case any proxy configurations I haven't thought about ruin the day.

I also hadn't thought too much about what happens when people host isso under a subfolder, e.g. `https://mysite.com/isso`, but this PR only checks for URL scheme and nothing more.

ix5:
Added tests

```
nosetests --with-doctest --with-coverage --cover-package=isso --cover-html isso/
.................................................................................
Name                        Stmts   Miss  Cover
-----------------------------------------------
isso/__init__.py              164     61    63%
isso/config.py                 77     11    86%
isso/core.py                   68     20    71%
isso/db/__init__.py            66      6    91%
isso/db/comments.py           135     20    85%
isso/db/preferences.py         17      0   100%
isso/db/spam.py                36      0   100%
isso/db/threads.py             15      1    93%
isso/dispatch.py               39     23    41%
isso/ext/__init__.py           10      0   100%
isso/ext/notifications.py     150    100    33%
isso/migrate.py               201     27    87%
isso/run.py                     5      0   100%
isso/utils/__init__.py         75     14    81%
isso/utils/hash.py             67      8    88%
isso/utils/html.py             50      1    98%
isso/utils/http.py             39     25    36%
isso/utils/parse.py            43      2    95%
isso/views/__init__.py         36      8    78%
isso/views/comments.py        468     95    80%
isso/wsgi.py                  103     26    75%
-----------------------------------------------
TOTAL                        1864    448    76%
----------------------------------------------------------------------
Ran 81 tests in 5.199s

OK
```

isso/views/comments.py
```diff
@@ -348,6 +347,23 @@ def _remote_addr(self, request):
...
+        """
+        Setting cookies to SameSite=None requires "Secure" attribute.
+        For http-only, we need to override the dump_cookie() default SameSite=None
+        or the cookie will be rejected.
+        https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Set-Cookie/SameSite#samesitenone_requires_secure
+        """
+        isso_host_script = self.isso.conf.get("server", "public-endpoint") or local.host
+        if isso_host_script.startswith("https://"):
+            secure = True
+            samesite = "None"
```
Lucas-C, line 360: Why adding this attribute? Shouldn't the default "Lax" policy be good enough?

Lucas-C:
The PR code & tests look good to me. Good job!

However I am a bit worried about "hardcoding" `SameSite=None` for all HTTPS endpoints:
wouldn't some users want to set it to `Strict`?

While I agree that this should be the default behaviour,
a suggestion in the discussion at https://github.com/posativ/isso/issues/682 was to make it optionnaly configurable,
and that sounds like a good idea to me.

ix5:
I've added a config option as suggested by @Lucas-C and @stefangehn and updated tests.

ix5:
@Lucas-C @blatinier rebased, anything else you'd like to see changed or prevents you from merging?

blatinier (approved):

ix5:
Sweet, thanks!
Just a heads up, tests are currently failing due to flakes complaining about unused imports

PR #702, opened 2021-02-05
Remove legacy werkzeug support

ix5:
Since 3e3ee8b, werkzeug is required in version 1.0 or above

Also align deps in `tox.ini` with the ones in `setup.py`


ix5:
Rebased and ready for review

jelmer (approved):

PR #703, opened 2021-02-05
docs: install: Reorder Install from source, py3.5

ix5:
Reorder steps so that js files are generated before `python setup.py develop`.

Clarify that isso should be installed in editable mode when installed from source.
Clarify that sphinx is needed for docs, and only for docs.
Clarify that npm is definitely required.

Isso requires python 3.5 now.

### DIFF
--- a/docs/docs/install.rst
+++ b/docs/docs/install.rst
@@ -79,7 +79,7 @@ Install from PyPi
 Requirements
 ^^^^^^^^^^^^
 
-- Python 2.7 or 3.4+ (+ devel headers)
+- Python 3.5+ (+ devel headers)
 - SQLite 3.3.8 or later
 - a working C compiler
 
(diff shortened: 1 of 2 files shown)


docs/docs/install.rst
```diff
@@ -186,24 +187,34 @@ To create a virtual environment (recommended), run:
...
-Build JavaScript frontend code:
+Install Isso and its dependencies:
 
 .. code-block:: sh
 
-    ~> make js
+    ~> python setup.py develop  # or `pip install -e .`
+    ~> isso run
+
+**Note:** Do not use ``python setup.py install``! That would only install over
```
jelmer, line 209: I don't follow this; this does include generated files (including locally changed ones), and I use it regularly.

ix5: Unfortunate wording on my part.
If isso is installed using `python setup.py install`, the isso artifact files will be copied to `<venv>/lib/python3.9/site-packages/isso-0.12.4-py3.9.egg/isso/`, but any changes you make from now on will not be carried over to the site-packages isso, unless you run the install command again.

Maybe others have better suggestions for the wording? What I want to strongly communicate is that the editable install option will save many from annoying misunderstandings and outdated files.

jelmer: I think this is quite subjective - I don't use venv, and "python install" is how I deploy isso. Perhaps just drop this note altogether? Alternative, just add a caution here about "setup.py install" and venv.

stefangehn: I agree that using `pip install -e .` is handy during development since you do not need to install the package again after every change. Since this part of the documentation is about installing from source I think both installation variants are fine and it depends on the use case which one to choose.

So all in all I'd say drop the final `Note: ` line, the rest is just fine.

jelmer (changes requested):
See earlier comments.

docs/docs/install.rst
ix5: Line dropped

ix5:
Pushed with requested changes,

jelmer (approved):

PR #704, opened 2021-02-07
Serve isso.css separately instead of inline

ix5:
Instead of embedding isso.css inside the client javascript (which required an `style-src: unsafe-inline` CSP), fetch isso.css from `api.endpoint+"/css/isso.css"`.

Allow clients to override fetch location using `data-isso-css-url="https://comments.example.org/css/isso.css"`

Note: No modification needed for packaging since isso.css is already included via MANIFEST.in.

Fixes https://github.com/posativ/isso/issues/584

### DIFF
--- a/docs/docs/configuration/client.rst
+++ b/docs/docs/configuration/client.rst
@@ -9,6 +9,7 @@ preferably in the script tag which embeds the JS:
     <script data-isso="/prefix/"
             data-isso-id="thread-id"
             data-isso-css="true"
+            data-isso-css-url="null"
             data-isso-lang="ru"
             data-isso-reply-to-self="false"
             data-isso-require-author="false"
@@ -96,6 +97,16 @@ override the API location:
 .. code-block:: html
 
     <script data-isso="/isso" src="/path/to/embed.min.js"></script>
+
+data-isso-css-url
+-----------------
+
+Set URL from which to fetch ``isso.css``, e.g. from a CDN.
+Defaults to fetching from the API endpoint.
+
+.. code-block:: html
+
+    <script src="..." data-isso-css-url="/path/to/isso.css"></script>
 
 data-isso-css
 -------------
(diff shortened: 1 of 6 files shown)


jelmer:
Doesn't this add an extra roundtrip?

ix5:
> Doesn't this add an extra roundtrip?

It sure does. All these "security" features cost a lot of requests, same as for CORS preflight :(

vincentbernat:
I serve JS from a CDN, not from the endpoint. It would be nice to make this configurable, but otherwise, no big trouble for me, as I also build a custom version (notably to remove languages I don't use to minimize size).

ix5:
> I serve JS from a CDN, not from the endpoint. It would be nice to make this configurable, but otherwise, no big trouble for me, as I also build a custom version (notably to remove languages I don't use to minimize size).

Added `data-isso-css-url` so you can serve your styles from a CDN. Nice site btw, Mr MTU Ninja ;)

docs/docs/configuration/client.rst
```diff
@@ -100,6 +101,16 @@ override the API location:
...
 -------------
 
+Set URL from which to fetch ``isso.css``, e.g. from a CDN.
+Defaults to fetching from the API endpoint.
+
+.. code-block:: html
+
+    <script src="..." data-isso-css-url="/path/to/isso.css"></script>
+
+data-isso-css-url
```
MorrisJobke, line 111: I guess this block was put at the wrong place. The text above this heading is for the attribute data-isso-css-url. Mind to move it below?

ix5: Thanks for catching this! Fixed.

ix5:
Rebased, @jelmer @vincentbernat @MorrisJobke any other suggestions?

blatinier (approved):

blatinier:
Everything is good there.
As for the added round trip it's not too bothersome since http/2 avoid creating multiple connexions.

Issue #566, opened 2019-08-14
Links and Threads Not Working


joshcsullivan:
Hello,

I've had Isso running for a few months now and love it, but I have one issue that I cannot seem to solve myself. My blog, https://blog.joshsullivan.io/ is hosted on https://blot.im/ a SSG, so I do not host the blog on a server myself and therefore Isso cannot be served as blog.joshsullivan.io/isso/. I instead created an Isso instance on a VPS at isso.joshsullivan.io but now links in Isso Admin point to https://isso.joshsullivan.io/correct-ending-of-url and the threads in Isso admin don't work at all, I assume because it is trying to find the comment path at the isso.joshsullivan.io address instead of the blog.joshsullivan.io.

Is there any way to correct this so threads and links in the admin panel work correctly?

I've set the hosts to

```
https://blog.joshsullivan.io/
http://blog.joshsullivan.io/
https://isso.joshsullivan.io/
http://isso.joshsullivan.io/
```

Cheers,
Josh

ix5:
At the moment, isso only stores *relative* URLs for threads - it is not multi-site-aware.

The client only sends the `pathname`, see [api.js](https://github.com/posativ/isso/blob/78997f491044b7d694ac7170edc32030544095b7/isso/js/app/api.js#L6).

If you want to change that behaviour, you will have to account for people already using isso:
Either, make a migration script for existing entries in the database to use the full URL, or write a clause for both cases, something like this (pseudo-code):

client:
```
host = window.location.host;
pathname = window.location.pathname;
```

Example thread for this purpose: `blog.joshsullivan.io/my-thread`
Hosts in config: `https://blog.joshsullivan.io/`, `http://blog.joshsullivan.io/`

server:
```
def sanitize_host(hostname):
    return hostname.lstrip("http://").lstrip("https://").rstrip("[\w]/")

hosts = list(set(map(sanitize_host, hosts)))

thread_host = get_from_request_params("host")
thread_uri = get_from_request_params("pathname")

if get_from_db(thread=thread_uri, host=thread_host):
    return_comments_and_stuff

# Legacy
for host in hosts:
    if get_from_db(thread=thread_uri, host=host):
        return_comments_and_stuff()
        break
```

ix5:
This is the workaround I use for now
https://github.com/ix5/isso/commit/bc6f0a0cf24f3ef038ee5a14b837d76e065b677a

With a monkey-patched `embed.min.js`: 
```
[...]

PR #711, opened 2021-03-01
ix5 workaround for multi-site awareness (full uris)

EchedeyLR:
See https://github.com/posativ/isso/issues/566 and https://github.com/ix5/isso/commit/bc6f0a0cf24f3ef038ee5a14b837d76e065b677a

### DIFF
--- a/isso/js/app/api.js
+++ b/isso/js/app/api.js
@@ -3,7 +3,7 @@ define(["app/lib/promise", "app/globals"], function(Q, globals) {
     "use strict";
 
     var salt = "Eech7co8Ohloopo9Ol6baimi",
-        location = function() { return window.location.pathname };
+        location = function() { return window.location.host + window.location.pathname };
 
     var script, endpoint,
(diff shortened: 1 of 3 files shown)


ix5:
This will clash with existing databases. You're supposed to use one isso instance per site as described in [docs: Multiple Sites](https://posativ.org/isso/docs/setup/multiple-sites/).

EchedeyLR:
My issue was only because I use a different domain for the isso instance. To be exact I don't know how to solve this issue since I don't understand the structure of Isso at all.

It cost me a lot even to find the reference you made to the issue in the github thread I mentioned because seems not so much people use Isso in this way.

If there is other way to solve the issue, I am open to solutions.

ix5:
As this will clash with existing setups (comments created before this change will stop showing up entirely), this change can not be accepted.

I'm open to suggestions that make Isso use  full URIs (minus the protocol `http`/`https`) internally, with an automatic database migration from older versions. Let's keep discussing in #566.

PR #733, opened 2021-07-01
Fix GitHub actions

BookGin:
This PR fixed the dependency issues that causes CI test to fail.

Unless isso explicitly supports Windows and PyPy3, I don't think we should include them in the CI test.
Also we should consider drop the support for Python 3.5 [which reached end-of-life](https://www.python.org/downloads/release/python-3510/).

### DIFF
--- a/.github/workflows/pythonpackage.yml
+++ b/.github/workflows/pythonpackage.yml
@@ -13,20 +13,21 @@ jobs:
       fail-fast: false
 
     steps:
-    - uses: actions/checkout@v2
-    - name: Set up Python ${{ matrix.python-version }}
+    - name: Check out repository code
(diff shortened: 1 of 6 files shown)


jelmer (approved):

jelmer:
Dropping support for Pytohn 3.5 LGTM. 

I think we should support pypy3; perhaps we can just ignore the test results from pypy3 for now?

BookGin:
Sure, how about Windows, MacOS and Python 3.9? (See the comment above.)

BookGin:
I've included Python 3.9. For dropping support for Python 3.5. I'll create a another PR.

jelmer (approved):

Issue #755, opened 2021-12-31
Python test coverage needs improvement


jelmer:
The test coverage for isso's Python code is very spotty; in order to be confident when making changes and adding features, we should make sure that all code at least has some degree of coverage.

ix5:
To elaborate, the python unit tests are in [isso/tests/](https://github.com/posativ/isso/tree/master/isso/tests).

To run Isso's unit tests, you can use `python -m unittest discover isso/tests` (or `pytest` installed via `pip install pytest`).

We should probably update the docs that still recommend using `nose` since it won't run on current python versions and is abandoned by its creators.

Guts:
Maybe we could start by add a linter like flake8 to improve code reliability?
We could also add it as a git hook through pr-commit to enforce code homogeneity and contributing guidelines?

jelmer:
Code linting and enforcing style seems like a good idea (and I'd welcome a PR to add that), but it is unrelated to better tests.

Guts:
>  but it is unrelated to better tests.

Woo @jelmer, you are quite harsh! 
I see a static analysis tool (linter) as a good complementary first step to better tests since it aims at improving software quality.

jelmer:
My apologies, I didn't mean to come across as harsh. I'm not disagreeing with you about the usefulness of a linter in general and its potential contributions to code quality. But they're orthogonal, and a linter won't give us significantly more confidence merging new PRs.

Guts:
No worries, I get it that we agree.

I've pushed a PR.

ix5:
> Maybe we could start by add a linter like flake8 to improve code reliability?

@Guts we already have flake8 (via `make flakes`), would you care to modify the config (inside `setup.cfg` so far) so that it catches more things?

ix5:
As a reference point, this is the current coverage:
```sh
$ pytest --doctest-modules --cov=isso isso/
======================= test session starts ===============
platform linux -- Python 3.10.3, pytest-7.1.1, pluggy-1.0.0
rootdir: /home/foo/isso/isso
plugins: cov-3.0.0
collected 87 items

isso/config.py ..                             [  2%]
isso/tests/test_comments.py ........          [ 51%]
isso/tests/test_config.py .                   [ 52%]
isso/tests/test_cors.py ..                    [ 55%]
isso/tests/test_db.py ...                     [ 58%]
isso/tests/test_guard.py .....                [ 64%]
isso/tests/test_html.py ........              [ 73%]
[...]

PR #756, opened 2022-01-05
Add pre-commit configuration

Guts:
**In this PR**

- add classical git hooks through [pre-commit](https://pre-commit.com/) to prevent "simple" issues when pushing code
- add basic flake8 git hook, reusing configuration from setup.cfg and selecting most important flags
- add specific section for https://pre-commit.ci/ - recommended because it automates hook updates and quick fixes when it's possible. This requires that a member with write permissions registers the repository on the application.
- add required as extra dependencies into setup.py for development

**Goal**

A little step to improve code reliability.

**Related issues**

Related to #755

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,0 +1,39 @@
+exclude: "node_modules|.venv|isso/tests"
+
+repos:
+  - repo: meta
+    hooks:
+      - id: check-hooks-apply
+
+  - repo: https://github.com/pre-commit/pre-commit-hooks
+    rev: v4.1.0
+    hooks:
+      - id: check-added-large-files
+        args: ['--maxkb=500']
+      - id: check-executables-have-shebangs
+      - id: check-case-conflict
+      - id: check-yaml
+        args: [--unsafe]
+      - id: detect-private-key
+      - id: end-of-file-fixer
+      - id: fix-encoding-pragma
+      - id: trailing-whitespace
+        args: [--markdown-linebreak-ext=md]
+
+  - repo: https://github.com/pycqa/flake8
+    rev: 4.0.1
+    hooks:
+      - id: flake8
+        language: python
+        files: ^isso/.*\.py$
+        types: [python]
+        args: ["--config=setup.cfg", "--select=E9,F63,F7,F82"]
+
+# Specific part for https://pre-commit.ci
+ci:
+  autofix_prs: true
+  autoupdate_schedule: quarterly
+  skip: [
+    "check-hooks-apply"
+  ]
+  submodules: false
--- a/setup.py
+++ b/setup.py
@@ -47,8 +47,8 @@ setup(
     install_requires=requires,
     extras_require={
         "dev": [
-                    "pre-commit>=2.15,<2.18",
-                    "flake8>4,<4.1",
+            "pre-commit>=2.15,<2.18",
+            "flake8>4,<4.1",
         ],
     },
     entry_points={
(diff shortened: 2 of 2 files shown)


Guts:
ping @jelmer and @ix5

jelmer (approved):

.pre-commit-config.yaml
```diff
@@ -0,0 +1,39 @@
...
+        args: [--markdown-linebreak-ext=md]
+
+  - repo: https://github.com/pycqa/flake8
+    rev: 4.0.1
+    hooks:
+      - id: flake8
+        language: python
+        files: ^isso/.*\.py$
+        types: [python]
+        args: ["--config=setup.cfg", "--select=E9,F63,F7,F82"]
```
jelmer, line 30: isn't this severely limiting what errors we'll get?

Guts: It's not to be too frustrating at the time of the commit or a PR.
If you prefer, we can enlarge this selection or apply everything.

.pre-commit-config.yaml
```diff
@@ -0,0 +1,39 @@
...
+
+# Specific part for https://pre-commit.ci
```
ix5, line 32: Am I understanding this correctly that there'll be a little bot auto-committing to this repo? I'd rather avoid this. Proposing changes is fine, but commit rights seem a bit overkill.
Or will this only auto-commit what it thinks are fixes to the branches of PR authors which we can then squash?

ix5 (commented):
I like your effort, @Guts. This certainly seems like something that could help catch little slip-ups in a fast-paced development environment.
I've left a few comments in the code.

In general though, I think running linting tools should be up to the discretion of individual committers. We don't want to force them to install _yet another_ framework on their machine (with auto-downloaded hook repos from a to their eyes untrusted Github source).

I'm perfectly fine with _recommending_ running flake8 before committing. flake8 is being run as CI on PRs already.

I'm also a bit confused because `pre-commit` seems to be two things at the same time:
- A framework for running "hooks" from its GitHub hook repository (which needs to be updated regularly) by creating a virtualenv (which I can't find any info about)
- A CI cloud service (free?) which requires push access for its bot to contributors' branches to push what it thinks are trivial fixes

The first one, as I've stated above, should in my opinion reside with individual contributors. We shouldn't force anyone to use it, but having the config for people who might want it seems a good idea.
The second one seems a bit intrusive. It's another unknown actor to trust. For my taste, the whole "sprinkle some packages on it" vibe of nodejs is not something I'd want to bring to this project. But then again, I'd not be entirely opposed to it. Waiting for other maintainers' opinions on this.

Don't get me wrong, I'm very much grateful for your efforts @Guts, but it seems this project will soon be 90% configs for some type of devops service while actual, structural or architectural issues are left hanging. If this helps us write unit tests or invites more contributors, that's a positive outcome. But as they say, adding fancy new things is easy, fixing bugs is hard ;)

jelmer:
> In general though, I think running linting tools should be up to the discretion of individual committers. We don't want to force them to install _yet another_ framework on their machine (with auto-downloaded hook repos from a to their eyes untrusted Github source).
> 
> I'm perfectly fine with _recommending_ running flake8 before committing. flake8 is being run as CI on PRs already.
Hmm, I'd forgotten about that. Adding flake8 to the "dev" dependencies seems reasonable. I'm not sure that the pre-commit config really adds much in that case.

-1 on automatic updates by bot from me as well.

Jelmer

PR #757, opened 2022-01-07
Avoid TypeError when #isso-root is not in the DOM

### LINKED ISSUE #306: BUG: Uncaught TypeError: Cannot set property 'textContent' of null

zaygraveyard:
`$(...)` returns `null` when the selector matches no element, and thus `$(...).length == 0` should not be used.

Fixes #306

### DIFF
--- a/isso/js/embed.js
+++ b/isso/js/embed.js
@@ -48,7 +48,7 @@ require(["app/lib/ready", "app/config", "app/i18n", "app/api", "app/isso", "app/
 
     function fetchComments() {
 
-        if ($('#isso-root').length == 0) {
+        if (!$('#isso-root')) {
             return;
         }
 


ix5 (approved):
Looks good to me and I can't see any immediate breakage. 9161983a was the commit that originally introduced this. I don't really see the necessity of this change for most users and am not qualified to weigh in on SPA/dynamic website functionality, but the syntax is definitely cleaner this way.

I'll wait a bit to give other people with similarly exotic setups to weigh in and then merge.

---

Just a few nits for writing better commit messages:

If you write `Fixes #306` on a separate line in your commit message, the linked issue will automatically be closed once this PR is merged.

Try to keep your commit title under 50 characters and the body text to a width of 80 characters. [Here's a little explanation](https://dev.to/noelworden/improving-your-commit-message-with-the-50-72-rule-3g79).

ix5:
@zaygraveyard It seems no one has brought any objections forward so far.
Would you mind changing the commit message to include the part that was changed?
Something like:
```
embed.js: Make check for isso-root more robust

<describe previous behaviour here>

<This fixes TypeError ...>

Fixes #306
```

zaygraveyard:
@ix5 Sure no problem :smile:

zaygraveyard:
@ix5 Is the new commit message better?

ix5:
Great, thanks! I merged your improvement.

Issue #763, opened 2022-01-27
configparser.InterpolationSyntaxError: for SMTP password


Aerya:
Hi,

How to allow Isso to parse pwds correctly?
`configparser.InterpolationSyntaxError: '%' must be followed by '%' or '(', found: '%!...@c...!Cjq'`

ix5:
That's an interesting bug you have found in Isso. Thanks for reporting it!

The issue is that the [`configparser.ConfigParser`](https://docs.python.org/3/library/configparser.html#configparser.ConfigParser) from Python that Isso uses tries to "interpret" (=interpolate) any string containing the `%` symbol using standard python string interpolation.

```
$ python3
>>> passwd = 'mysecretpw%s'
>>> passwd % 'hidden'
'mysecretpwhidden'
>>> passwd = 'mysecretpw%(mystr)s'
>>> passwd % {'mystr': 'hidden'}
'mysecretpwhidden'
```

The solution for you is to (temporarily) stop using passwords that contain `%` (or use `%%` in the config, `%` in the web interface). The next version of Isso should have this fixed.

Aerya:
Thanks for incoming patch. I'll wait for it, I won't use less secure pwds.

PR #764, opened 2022-01-27
config: Disable interpolation of values

ix5:
Do not attempt to use python string interpolation on values.
This is especially important for parsing passwords that might contain '%'

<strike>Strip single and double quotes from values. This might help with people who try to supply their admin password surrounded in quotes.</strike> (Edit: removed)

See: https://docs.python.org/3.9/library/configparser.html#configparser.ConfigParser

Also adjust tests accordingly and add new test case for strings which should not be interpolated.

This avoids errors such as the following:
```
configparser.InterpolationSyntaxError:
'%' must be followed by '%' or '(', found: '%!...@c...!Cjq'
```

Fixes https://github.com/posativ/isso/issues/763

---
I also came across this while looking through the code:

treewide: Remove remaining py2 u-String usage
 
In python3, every string is a unicode string by default.
There is no need to keep the prefixed `u"mystring"`-style.

### DIFF
--- a/isso/config.py
+++ b/isso/config.py
@@ -78,9 +78,21 @@ class Section(object):
 class IssoParser(ConfigParser):
     """Parse INI-style configuration with some modifications for Isso.
 
-        * parse human-readable timedelta such as "15m" as "15 minutes"
-        * handle indented lines as "lists"
+        * Parse human-readable timedelta such as "15m" as "15 minutes"
+        * Handle indented lines as "lists"
+        * Disable string interpolation ('%s') for values
     """
+
+    def __init__(self, *args, **kwargs):
+        """Supply modified defaults to configparser.ConfigParser
+        https://docs.python.org/3.9/library/configparser.html#configparser.ConfigParser
+        allow_no_value=True: Allow empty config keys
+        interpolation=None:  Do not attempt to use python string interpolation on
+                             values. This is especially important for parsing
+                             passwords that might contain '%'
+        """
+        return super(IssoParser, self).__init__(
+            allow_no_value=True, interpolation=None, *args, **kwargs)
 
     def getint(self, section, key):
         try:
@@ -107,7 +119,7 @@ class IssoParser(ConfigParser):
 
 def new(options=None):
 
-    cp = IssoParser(allow_no_value=True)
+    cp = IssoParser()
 
     if options:
         cp.read_dict(options)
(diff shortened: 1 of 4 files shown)


isso/config.py
```diff
@@ -107,7 +125,7 @@ def section(self, section):
 
 def new(options=None):
 
-    cp = IssoParser(allow_no_value=True)
+    cp = IssoParser()
```
ix5, line 128: Note that I moved the `allow_no_value` arg into the `__init__` function to make it easier to keep things in sync. Otherwise you'd have to adjust the tests as well every time `IssoParser` needs new args.

ix5:
Pinging @stefangehn @jelmer in case you have different stylistic preferences.
Treating quotes differently might warrant an entry into CHANGES as well.

isso/config.py
```diff
@@ -78,10 +78,28 @@ def getboolean(self, key):
...
+        allow_no_value=True: Allow empty config keys
+        interpolation=None:  Do not attempt to use python string interpolation on
+                             values. This is especially important for parsing
+                             passwords that might contain '%'
+        """
+        return super(IssoParser, self).__init__(
+            allow_no_value=True, interpolation=None, *args, **kwargs)
+
+    def get(self, section, key, *args, **kwargs):
+        # Remove leading and trailing quotes from values
```
stefangehn, line 99: Not a fan of this since it's not strictly needed any actually disallows passwords starting or ending with these characters (in addition to whitespace that may actually be gobbled up by the ini parser)

isso/config.py
```diff
@@ -78,10 +78,28 @@ def getboolean(self, key):
...
+        return super(IssoParser, self).get(section, key, *args, **kwargs)\
+            .strip("'").strip('"')
```
stefangehn, line 101: strip can be passed multiple chars to strip off a string, no need to call it twice

stefangehn (commented):
As mentioned in-line I wouldn't strip quotes, disabling interpolation is certainly a good idea though.

isso/config.py
jelmer: +1

isso/config.py
jelmer: Why do we need to strip these? It seems like it would make it impossible to actually use these characters at the beginning and end of strings.

jelmer (changes requested):
setting allow_no-value=True and interpolation=None seems reasonable to me, but please see other comments

isso/config.py
ix5: Alright, thank you for your input! Adjusted accordingly.

isso/config.py
ix5: See above.

PR #768, opened 2022-02-08
fix comment in config example

kevinnls:
since code blocks don't unwrap lines, part of comment in config example is rendered as non-comment.

added `;` to start of line to fix that.


ix5:
@kevinnls would you mind writing a more precise commit message?

You already shared why this patch is needed in the PR description, please include it inside the commit message as well.

The title should reference the part of Isso that is affected, e.g. `docs: quickstart: Fix comment in config example`

I know it seems a bit overkill for such a small patch, but consistency is key.

kevinnls:
oops, my bad. apologies for the lazy. i was doing in gh's web UI. will amend shortly

kevinnls:
oh. that was awkward.

kevinnls:
sorry about that @ix5 new PR or would it be fine with you to reopen and merge? i don't have permission to reopen apparently

kevinnls:
i'll open a new pr. sorry

PR #769, opened 2022-02-08
docs: quickstart: fix comment in sample config

kevinnls:
redo #768 

since code blocks don't unwrap lines, part of comment in config example is rendered as non-comment.

added `;` to start of line to fix that.

### DIFF
--- a/docs/docs/quickstart.rst
+++ b/docs/docs/quickstart.rst
@@ -19,7 +19,7 @@ sane defaults.
 
     [general]
     ; database location, check permissions, automatically created if it
-    does not exist
+    ; does not exist
     dbpath = /var/lib/isso/comments.db
     ; your website or blog (not the location of Isso!)
     host = http://example.tld/


ix5:
Please do not delete your branches, there's no need for that. That way, you can re-use an existing PR.

What I meant for you to do was to add the following from your PR message to the commit message:
> Since code blocks don't unwrap lines, part of comment in config example is rendered as non-comment.
> Added ";" to start of line to fix that.

kevinnls:
i hadn't actually deleted. i force amended the commit message and force pushed a grafted branch and messed up the history ://

i fixed it on the same branch, but for some reason the [file count and diff](https://github.com/posativ/isso/pull/768/files) of the previous PR wasn't updating. might've been because it was closed?

> What I meant for you to do was to add the following from your PR message to the commit message:

ah. didn't realise. thank you 😄

ix5:
Not quite sure, it seems GitHub will automatically force-close a PR if no common ancestors are found on the pushed branch.
In any case, you'll be safer by using a full clone instead of shallow+grafted complications, but I can understand wanting to save bandwidth.

In any case, thank you for your contribution.

PR #793, opened 2022-02-13
[UGLY] admin: Use textarea for comment editing

ix5:
Use a `<textarea>` element to let the user properly edit comments, where `contenteditable` was poorly suited for multi-line input.

---

If anyone has a cleaner solution, go for it.
This is my stopgap solution to make comment editing in the admin work again (which got bungled by the combination of https://github.com/posativ/isso/pull/470 and https://github.com/posativ/isso/pull/604, where currently the edited content is wrapped in unwanted literal `<pre>` tags)

### DIFF
--- a/isso/css/admin.css
+++ b/isso/css/admin.css
@@ -135,3 +135,10 @@ a {
 .hidden {
   display: none;
 }
+.text-wrapper .text {
+    margin: 0.5em 0;
+}
+.text-wrapper .text textarea {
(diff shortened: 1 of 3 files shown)


ix5:
This has been open for 10 days now. As no one has objected or put forward a better solution, I'll merge this in the interest of at least having a functional admin interface again, however ugly the implementation may be under the hood.

Issue #754, opened 2021-12-31
Improve JavaScript code test coverage


jelmer:
There is barely any test coverage for the javascript code at the moment. It would be great to at least have some basic tests to validate that the code is valid, and ideally also that it makes the right HTTP requests.

**Edit:** (addded by @ix5)
### Related PRs
- https://github.com/posativ/isso/pull/800
- https://github.com/posativ/isso/pull/822
- https://github.com/posativ/isso/pull/801
- https://github.com/posativ/isso/pull/807
- https://github.com/posativ/isso/pull/826
- https://github.com/posativ/isso/pull/868
- https://github.com/posativ/isso/pull/884
- https://github.com/posativ/isso/pull/889
- https://github.com/posativ/isso/pull/890

ix5:
If someone experienced in JS would like to propose a solution, go for it!

@blatinier you seemed to be interested in migrating (some) of the client code to VanillaJS or something similar - let's start by covering the existing code ;)

ix5:
@stefangehn @vincentbernat @pellenilsson @rocka @Lucas-C pinging you for recommendations/insights into JS testing.

Baring any rewrite that gets rid of RequireJS/almond, the testing framework/solution/... should be compatible with these technologies. I'll throw [jest](https://jestjs.io/) into the room, not having any particular experience with it but it looks well-supported.

Lucas-C:
I can confirm that, in my experience, `jest` is very good!

ix5:
> I can confirm that, in my experience, `jest` is very good!

That's great to hear! Do you have any particular examples of similar usage which might help in implementing testing for Isso?

In any case, we'd also have a need of bringing up the server part (via docker?) to test out responses (or find a way to mock them).

Lucas-C:
> Do you have any particular examples of similar usage which might help in implementing testing for Isso?

No I don't have any useful example at hand...

However, what would you think about also introducing a code linter as part of the GitHub Actions pipeline?
`eslint` for example. Maybe I could find the time to submit a PR introducing it...

ix5:
> No I don't have any useful example at hand...

No worries, maybe someone else has.
 
> However, what would you think about also introducing a code linter as part of the GitHub Actions pipeline? `eslint` for example. Maybe I could find the time to submit a PR introducing it...

[...]

PR #800, opened 2022-02-19
[MAJOR] js: Move to webpack, pug from r.js, jade

ix5:
# Introduction

As already discussed in https://github.com/posativ/isso/issues/754, try to make Isso more maintainable by facilitating modern testing practices.

This PR is a bit massive, but trying to make `pug` or other updated deps play nice with RequireJs just for the sake of a clean history is not something I'm prepared to do.

So, here goes:

---

### js: Move to webpack, pug from r.js, jade

In the interest of maintainability and integration with other testing tools, get rid of `RequireJS` (sometimes also called `r.js` or `requirejs`)

This necessitates the following changes:
- Removing the `almond` (`AMD`) module loader (superseded by webpack internals)
- Removing `requirejs-text` for loading static resources (superseded by webpack `asset/source` loader)
- Removing `jade` templating engine (superseded by `pug`, which represents the original `jade` project that had to be renamed due to trademark issues)
  `pug` v2+ no longer accepts `'mystr-{#foo.bar}'`-style variables inside attributes, so `*.pug` templates need to be changed accordingly (`mystr' + foo.bar`)
  Note: templates have been moved to `templates/` directory and renamed from `.jade` to `.pug`.
- `jade-runtime` and accompanying shims (superseded by simply `pug-runtime`)
- Move `.svg` files inside own dir, simplify loading via `webpack`'s `asset/source`
- Add `webpack` config
- Add build scripts to `package.json`
- Adjust `Makefile` to use `webpack` via `npm run`
- Emit JS source map files from `webpack` (eases debugging)
- `.gitignore` JS source map files

While RequireJS utilized the following syntax:
```
define(["app/foo"], function(foo) {
  var bar = doSomeStuff...
  return bar;
})
```
Jest and other testing libraries need the NodeJS standard `CommonJS` style:
```
const foo = require("app/foo");
var bar = doSomeStuff...
module.exports = { bar };
```
Note: ES6/ECMAScript2015-style `import` has not been used since it would need transpilation via e.g. `babel` to be understood by Jest.

The RequireJS-style wrapping of all code inside `define({...})` caused everything to be indented by at least one level.
To keep the changes in this commit manageable, indentation will be removed in a later commit.

### isso: js: treewide: De-indent one level

Since switching from RequireJS to CommonJS syntax, the `define()` wrapper is no longer needed and as such, the the code formerly inside parentheses can be de-indented by one level.

### isso: js: i18: RequireJS to module.exports

(Just an extension of previous commit, to avoid touching too many files at once)

### demo: Use dev version of embed.js

This makes it easier to debug when testing. The demo shouldn't be user-facing, but either way, the performance penalty should be outweighed by debugging benefits.

### isso: js: Reduce deps for app/lib/
(Just a minor commit to make the following unit tests easier)

---

# Verifying

You can verify this commit by checking out this PR:
```sh
git clone https://github.com/posativ/isso isso
cd isso/
git fetch origin pull/800/head:pull-800
git checkout pull-800
```

And then installing everything as documented in [docs: Install from Source](https://posativ.org/isso/docs/install/#install-from-source).

Then, either embed Isso into your site or use the demo at [localhost:8080/demo](http://localhost:8080/demo)

---

A note for people from the future: You can use `git blame --ignore-rev=<sha>` to "ignore" the de-indent commits. See [A better git blame](https://www.michaelheap.com/git-ignore-rev/).

Note to myself: We can probably ditch the templates altogether and simply use plain html with javascript string interpolation. That way, we lose `pug`, `pug-loader` and `pug-runtime`, which would make the bundle significantly smaller.


ix5:
Pinging @stefangehn, @blatinier, @Lucas-C and anyone else interested in moving this project forward.

ix5:
I've created a package for easier testing:
[isso-0.12.5-webpack.tar.gz](https://github.com/posativ/isso/files/8127557/isso-0.12.5-webpack.tar.gz)

isso/js/app/lib/identicons.js
```diff
@@ -4,92 +4,91 @@
...
-                color = config["avatar-fg"][i % config["avatar-fg"].length];
+    svg.appendChild(rect);
+};
 
-            for (var x=0; x<Math.ceil(GRID/2); x++) {
-                for (var y=0; y<GRID; y++) {
+/**
+ * Pick random squares to fill in.
+ */
+var generateIdenticon = function(key, padding, size, config) {
```
Lucas-C, line 36: `config` should probably be  passed to this function on line 85 of this file

isso/js/app/templates/comment-loader.pug
```diff
@@ -1,2 +1,2 @@
-div(class='isso-comment-loader' id='isso-loader-#{comment.name}')
+div(class='isso-comment-loader' id='isso-loader-' + comment.name)
```
Lucas-C, line 1: Shouldn' `${}` be replaced on next line too?

isso/js/app/templates/comment.pug
```diff
@@ -1,21 +1,21 @@
...
-                a(class='author' href='#{comment.website}' rel='nofollow')
+                a(class='author' href=comment.website rel='nofollow')
                     = bool(comment.author) ? comment.author : i18n('comment-anonymous')
             else
                 span(class='author')
                     = bool(comment.author) ? comment.author : i18n('comment-anonymous')
             span(class="spacer") &bull;
-            a(class='permalink' href='#isso-#{comment.id}')
-                time(title='#{humanize(comment.created)}' datetime='#{datetime(comment.created)}')
+            a(class='permalink' href='#isso-' + comment.id)
```
Lucas-C, line 17: There are still usages of `${}` on lines 35-36-37 below in this file

isso/js/app/template.js
```diff
@@ -0,0 +1,93 @@
+const runtime = require("pug-runtime");
+const utils = require("app/utils");
+
+// TODO get rid of pug-loader entirely
```
Lucas-C, line 4: I'm not familiar with `pug`:
is the work in this file finished, or are there things that remain to be properly configured / cleaned up?

Lucas-C (approved):
This is great!
Awesome job on moving everything to ES6 modules & `webpack` O.O
Maybe this should be merged as part of #801 as we would have the added confidence of the new tests?

isso/js/app/lib/identicons.js
ix5: In this particular instance you are right, but looking over this code again, it occurred to me that Isso currently doesn't use the `blank`/`generateBlank` functions at all.

I'll adjust this commit to let `generateBlank` use a passed `config` for the sake of keeping current functionality.

isso/js/app/templates/comment-loader.pug
ix5: That's actually intended. Doing this is no allowed any more:
```
div(id='isso-loader-#{comment.name}')
```
See [pugjs: Attribute interpolation](https://pugjs.org/language/attributes.html#attribute-interpolation) and [pugjs: v2 migration](https://pugjs.org/api/migration-v2.html#attribute-interpolation)

However, anything outside parentheses is rendered as plain text if not using the `#{}` syntax.

This is `pug` being extremely stupid, imo, but I can't change that.

isso/js/app/templates/comment.pug
ix5: See above comment

isso/js/app/template.js
ix5: I should have clarified this: `pug-loader` allows one simplify loading the templates, loading them as "modules" right inside the `require` module loading syntax.

So one can do `require("pug!path/to/templates/xyz.pug")` instead of figuring out how to load templates from disk. This "loader" needs to be configured in `webpack`, as I have done.

However, using `pug-loader` pulls in _another_ dependency, and, most importantly, clashes with `Jest` testing (Jest is not able to recognize this loader syntax).

Will try to amend commit with this info.

ix5:
> This is great! Awesome job on moving everything to ES6 modules & `webpack`

Thank you for reviewing! Those questions you asked were all obstacles I stumbled upon while porting and was kind of too lazy to document. Thanks for keeping me honest ;) I'll document as I have written in individual responses inline.

> Maybe this should be merged as part of #801 as we would have the added confidence of the new tests?

While waiting for the test suite to be adequate might sound like a sensible strategy for keeping bugs out, it'll also mean keeping up a massive amount of rebase work and effectively blocking any contribs to the client (or doing a lot of merge conflict work).

So I'll be selfish and merge this `webpack`/`pug` change soon-ish, before tests are done. My "justification" for this is that once 0.12.6 is released, we can afford to be open for integrating more major changes.

isso/js/app/templates/comment-loader.pug
ix5: Turns out I was wrong here, using ["buffered code"](https://pugjs.org/language/code.html#buffered-code) is appropriate.

This:
```
a(class='reply' href='#') #{i18n('comment-reply')}
```
becomes this:
```
a(class='isso-reply' href='#')
    = i18n('comment-reply')
```

Note that the additional newline and indentation is needed.
This does not work:
```
a(class='isso-reply' href='#') = i18n('comment-reply')
```

isso/js/app/lib/identicons.js
ix5: Done

isso/js/app/templates/comment-loader.pug
ix5: Done

isso/js/app/template.js
ix5: Amended with comment.

PR #802, opened 2022-02-23
setup.py, README: Generate proper Project Description for PyPi

ix5:
### setup.py: Include README in pypi upload

This allows for more information by embedding the README at https://pypi.org/project/isso/ as a "Project Description".

### README: ~~Point LICENSE to current GH master,~~ shuffle sections a bit

~~Reasoning: Use absolute link to allow embedding the rendered README on pypi (where the relative-path LICENSE is not available).~~

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,9 +4,10 @@ Isso – *Ich schrei sonst* – is a lightweight commenting server written in
 Python and JavaScript. It aims to be a drop-in replacement for
 [Disqus](http://disqus.com).
 
-![Isso in Action](http://posativ.org/~tmp/isso-sample.png)
+See **[posativ.org/isso](http://posativ.org/isso/)** for more details and
+documentation.
 
-See [posativ.org/isso](http://posativ.org/isso/) for more details.
+![Isso in Action](http://posativ.org/~tmp/isso-sample.png)
 
 ## License
 MIT, see [LICENSE](LICENSE).
--- a/setup.py
+++ b/setup.py
@@ -1,11 +1,16 @@
 #!/usr/bin/env python3
 # -*- encoding: utf-8 -*-
 
+from pathlib import Path
 from setuptools import setup, find_packages
 
 requires = ['itsdangerous', 'Jinja2', 'misaka>=2.0,<3.0', 'html5lib',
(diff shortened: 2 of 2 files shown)


jelmer (approved):

README.md
```diff
@@ -9,7 +9,7 @@ Python and JavaScript. It aims to be a drop-in replacement for
 See [posativ.org/isso](http://posativ.org/isso/) for more details.
 
 ## License
-MIT, see [LICENSE](LICENSE).
+MIT, see [LICENSE](https://github.com/posativ/isso/blob/master/LICENSE).
```
jelmer, line 13: This breaks things for forks, can we just refer to the local file?

jelmer (changes requested):
See below.

README.md
ix5: You cannot link to package resources inside PyPi, even if you include them in `MANIFEST.in`.
(I tested via https://test.pypi.org/project/isso-testpypi/ and the link goes to a 404, as you'd expect: https://test.pypi.org/project/isso-testpypi/LICENSE )

jelmer:
On Wed, Feb 23, 2022 at 01:45:26PM -0800, ix5 wrote:
> @ix5 commented on this pull request.
> > @@ -9,7 +9,7 @@ Python and JavaScript. It aims to be a drop-in replacement for
>  See [posativ.org/isso](http://posativ.org/isso/) for more details.
>  
>  ## License
> -MIT, see [LICENSE](LICENSE).
> +MIT, see [LICENSE](https://github.com/posativ/isso/blob/master/LICENSE).
> 
> You cannot link to package resources inside PyPi, even if you include them in `MANIFEST.in`.
> (I tested via https://test.pypi.org/project/isso-testpypi/ and the link goes to a 404, as you'd expect: https://test.pypi.org/project/isso-testpypi/LICENSE )

I don't think we should bother fixing it then; ideally we'd just filter out the entire license section since the license is already shown on the left hand side on Pypi.

ix5:
> ideally we'd just filter out the entire license section

Done.

Personally, I think this is the ugliest code I've ever committed, but I agree with your sentiment. Unless someone comes up with a better/cleaner/more beautiful solution, I'm merging this within 2 days.

Issue #720, opened 2021-04-03
ISSO_CORS_ORIGIN doesn't appear to work as advertised


vsajip:
According to the documentation, you can use an `ISSO_CORS_ORIGIN` environment variable to set e.g. wildcards for the `Access-Control-Allow-Origin` CORS header. However, it doesn't work in practice, because the code that uses it is as shown in this fragment:

```python
    def func(environ):
        if 'ISSO_CORS_ORIGIN' in environ:
            return environ['ISSO_CORS_ORIGIN']
```
However, the `environ` passed into `func()` is a restricted WSGI environment (`request.environ`) rather than the full process environment (`os.environ`). Therefore, the code should use `os.environ` in the above fragment rather than just `environ`. Or is there some other way that an `ISSO_CORS_ORIGIN` OS environment variable is supposed to find its way into a `request.environ`?

ix5:
Using `ISSO_CORS_ORIGIN=*.example.tld` as written in isso's documentation is seemingly [against spec anyway, see this Stackoverflow answer](https://stackoverflow.com/a/14006226/2193463).

I've prepared a branch that rectifies isso's behaviour, see [ix5/isso@cors-from-os-environ](https://github.com/ix5/isso/tree/cors-from-os-environ) but that will still result in browsers refusing wildcard subdomains.

The correct way to do this would be to check if the originating URL matches the wildcard of `ISSO_CORS_ORIGIN` and then return a fully qualified URL (with everything after and including the slash chopped off).

Lucas-C:
This is strange...

On my side, it works fine:
```
$ cat /etc/systemd/system/isso.service
[Unit]
Description=isso comments daemon

[Service]
Environment=ISSO_CORS_ORIGIN=*.chezsoi.org
WorkingDirectory=/usr/share/nginx/html/lucas/isso
ExecStart=/usr/local/bin/pew in isso isso -c isso.cfg run
Restart=always

[Install]
WantedBy=multi-user.target

$ curl -v -o /dev/null 'https://chezsoi.org/lucas/isso/?uri=/lucas/blog/pages/bienvenue.html'
...
< access-control-allow-origin: http://chezsoi.org
```

Lucas-C:
@ix5 Maybe you could create a PR from your branch?

ix5:
## CORS

First off, a bit of explanation how CORS is supposed to work, since there might have been a few misunderstandings that lead to this.

[...]

PR #803, opened 2022-02-23
Remove ISSO_CORS_ORIGIN and references

ix5:
Reason for revert: This feature never worked.

See discussion in #720


ix5:
(including @Lucas-C and trying to find someone to convert explanations from #720 into docs)

Issue #121, opened 2014-08-13
 After commenting should the comment counter update immediately


dermariusz:

ix5:
Tracked in https://github.com/posativ/isso/pull/804

PR #804, opened 2022-02-24
[WIP] js: isso: Update comment counter on posting, deleting

ix5:
See https://github.com/posativ/isso/issues/121

This is a bit of a band-aid, so hold off with merging until work begins on 0.13 or find a better way, ideally via websockets to update comments dynamically.


ix5:
This might run into issues where they have been new comments posted by others (increasing the counter), but comments on page would still be pre-refresh. That'd lead to e.g. 5 reported comments when the page had 3 when the user visited it and himself added a fourth.

PR #805, opened 2022-02-26
views: moderation: Add unit tests, proper response instead of "Yo", apiDoc clarifications

fliiiix:
Things this PR addresses:
* Replace Yo response from moderation with a more appropriate response
* Add tests for moderation
* Add tests for unsubscribe and fix doc string

Contributes to #755

### DIFF
--- a/isso/tests/test_comments.py
+++ b/isso/tests/test_comments.py
@@ -660,6 +660,43 @@ class TestModeratedComments(unittest.TestCase):
         self.assertEqual(self.app.db.comments.get(id_), None)
 
 
+class TestUnsubscribe(unittest.TestCase):
+
+    def setUp(self):
+        fd, self.path = tempfile.mkstemp()
+        conf = config.load(
+            pkg_resources.resource_filename('isso', 'defaults.ini'))
+        conf.set("general", "dbpath", self.path)
+        conf.set("moderation", "enabled", "true")
+        conf.set("guard", "enabled", "off")
+        conf.set("hash", "algorithm", "none")
+
+        class App(Isso, core.Mixin):
+            pass
+
+        self.app = App(conf)
+        self.app.wsgi_app = FakeIP(self.app.wsgi_app, "192.168.1.1")
+        self.client = JSONClient(self.app, Response)
+
+        # add default comment
+        rv = self.client.post(
+            '/new?uri=test', data=json.dumps({"text": "..."}))
+        self.assertEqual(rv.status_code, 202)
+
+    def tearDown(self):
+        os.unlink(self.path)
+
+    def testUnsubscribe(self):
+        id_ = 1
+        email = "test@test.example"
+        key = self.app.sign(('unsubscribe', email))
+
+        # GET should return some html form
+        rv_unsubscribe_get = self.client.get('/id/%d/unsubscribe/%s/%s' % (id_, email, key))
+        self.assertEqual(rv_unsubscribe_get.status_code, 200)
+        self.assertIn(b"Successfully unsubscribed", rv_unsubscribe_get.data)
+
+
 class TestPurgeComments(unittest.TestCase):
 
     def setUp(self):
--- a/isso/tests/test_comments.py
+++ b/isso/tests/test_comments.py
@@ -621,7 +621,7 @@ class TestModeratedComments(unittest.TestCase):
         action = "activate"
         rv_activated = self.client.post('/id/%d/%s/%s' % (id_, action, signed))
         self.assertEqual(rv_activated.status_code, 200)
-        self.assertEqual(rv_activated.data, b"Yo")
+        self.assertEqual(rv_activated.data, b"Comment has been activated")
 
         # Activating should be idempotent
         rv_activated = self.client.post('/id/%d/%s/%s' % (id_, action, signed))
@@ -635,7 +635,7 @@ class TestModeratedComments(unittest.TestCase):
         action = "delete"
         rv_deleted = self.client.post('/id/%d/%s/%s' % (id_, action, signed))
         self.assertEqual(rv_deleted.status_code, 200)
-        self.assertEqual(rv_deleted.data, b"Yo")
+        self.assertEqual(rv_deleted.data, b"Comment has been deleted")
 
         # Comment should no longer exist
         self.assertEqual(self.app.db.comments.get(id_), None)
--- a/isso/tests/test_comments.py
+++ b/isso/tests/test_comments.py
@@ -696,6 +696,16 @@ class TestUnsubscribe(unittest.TestCase):
         self.assertEqual(rv_unsubscribe_get.status_code, 200)
         self.assertIn(b"Successfully unsubscribed", rv_unsubscribe_get.data)
 
+        # Incomplete key should fail
+        key = self.app.sign(['unsubscribe'])
+        rv_incomplete_key = self.client.get('/id/%d/unsubscribe/%s/%s' % (id_, email, key))
+        self.assertEqual(rv_incomplete_key.status_code, 403)
+
+        # Wrong key type should fail
+        key = self.app.sign(1)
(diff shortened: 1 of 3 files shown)


isso/tests/test_comments.py
```diff
@@ -604,6 +604,115 @@ def testAddComment(self):
...
+        rv = self.client.post(
+            '/new?uri=moderated', data=json.dumps({"text": "..."}))
+        self.assertEqual(rv.status_code, 202)
+        self.assertEqual(self.client.get('/id/1').status_code, 200)
+        self.assertEqual(self.app.db.comments.get(id_)["mode"], 2)
+        self.assertEqual(self.app.db.comments.get(id_)["text"], "...")
+
+        # GET should return some html form
+        action = "activate"
+        rv_activat_get = self.client.get('/id/%d/%s/%s' % (id_, action, signed))
```
ix5, line 622: `rv_activate_get`

isso/tests/test_comments.py
```diff
@@ -604,6 +604,115 @@ def testAddComment(self):
...
+        self.assertEqual(rv_activat_get.status_code, 200)
+        self.assertIn(b"Activate: Are you sure?", rv_activat_get.data)
+        self.assertIn(b"http://invalid.localmoderated#isso-1", rv_activat_get.data)
+
+        # GET should return some html form
```
ix5, line 627: Any reason for duplicating this test?

isso/tests/test_comments.py
```diff
@@ -604,6 +604,115 @@ def testAddComment(self):
...
         self.assertEqual(self.client.get('/?uri=test').status_code, 200)
 
+    def testModerateComment(self):
+
+        id_ = 1
+        signed = self.app.sign(id_)
+
+        # Create new comment, should have mode=2 (pending moderation)
+        rv = self.client.post(
+            '/new?uri=moderated', data=json.dumps({"text": "..."}))
```
ix5, line 614: Maybe use `uri=/moderated` (with slash) so that later on, the assert with `invalid.local/moderated` instead of `invalid.localmoderated` is not as confusing

isso/tests/test_comments.py
```diff
@@ -604,6 +604,115 @@ def testAddComment(self):
...
+        self.assertEqual(self.app.db.comments.get(id_)["mode"], 1)
+
+        # Edit comment
+        action = "edit"
+        rv_edit = self.client.post('/id/%d/%s/%s' % (id_, action, signed), data=json.dumps({"text": "new text"}))
+        self.assertEqual(rv_edit.status_code, 200)
+        self.assertEqual(json.loads(rv_edit.data)["id"], id_)
+        self.assertEqual(self.app.db.comments.get(id_)["text"], "new text")
+
+        # Wrong action on comment is handles by the routing
```
ix5, line 655: handled

isso/tests/test_comments.py
```diff
@@ -604,6 +604,115 @@ def testAddComment(self):
         self.app.db.comments.activate(1)
         self.assertEqual(self.client.get('/?uri=test').status_code, 200)
 
+    def testModerateComment(self):
+
+        id_ = 1
```
ix5, line 609: Using `id_` instead of my usage of private `_id` is much smarter, good catch!

isso/tests/test_comments.py
```diff
@@ -604,6 +604,115 @@ def testAddComment(self):
...
+        email = "test@test.example"
+        key = self.app.sign(('unsubscribe', email))
+
+        # GET should return some html form
+        rv_unsubscribe_get = self.client.get('/id/%d/unsubscribe/%s/%s' % (id_, email, key))
+        self.assertEqual(rv_unsubscribe_get.status_code, 200)
+        self.assertIn(b"Successfully unsubscribed", rv_unsubscribe_get.data)
+
+        # Incomplet key should fail
```
ix5, line 706: Incomplete, same for `rv_incomplete_key`

isso/views/comments.py
```diff
@@ -649,7 +646,7 @@ def unsubscribe(self, environ, request, id, email, key):
                 &lt;/script&gt;
 
     @apiSuccessExample Using POST:
-        Yo
+        Comment is deleted
```
ix5, line 649: "Comment deleted" / "Comment has been deleted" ?

isso/views/comments.py
```diff
@@ -689,7 +686,7 @@ def moderate(self, environ, request, id, action, key):
             with self.isso.lock:
                 self.comments.activate(id)
             self.signal("comments.activate", thread, item)
-            return Response("Yo", 200)
+            return Response("Comment is activated", 200)
```
ix5, line 689: "Comment activated" / "Comment has been activated" ?

isso/views/comments.py
```diff
@@ -704,7 +701,7 @@ def moderate(self, environ, request, id, action, key):
             self.cache.delete(
                 'hash', (item['email'] or item['remote_addr']).encode('utf-8'))
             self.signal("comments.delete", id)
-            return Response("Yo", 200)
+            return Response("Comment is deleted", 200)
```
ix5, line 704: dito

ix5 (changes requested):
Nice effort! Some minor nits here and there.

Please also keep commit message titles under 50 chars (or at least under 70, if otherwise not possible), else GitHub will hard-break the title with ellipses.

isso/tests/test_comments.py
fliiiix: its also what PEP8 tells you to do ;)

isso/tests/test_comments.py
fliiiix: nope this is a rebase mistake i overlooked

fliiiix:
Thanks for the review :tada:  i addressed all your feedback, take an other look if you spot any issues

Except the thing with the git commit message length, im not to sure what i should do there I can drop the prefix to make them shorter, but i guess the idea is to use prefixes? I can create super short unspecific commit messages which i feel would be worse. 
So i am a bit lost how i should solve that any ideas?

isso/views/comments.py
```diff
@@ -561,25 +561,19 @@ def delete(self, environ, request, id, key=None):
     @apiParam {string} email
         The email address of the subscriber.
     @apiParam {string} key
-        The key to authenticate the subscriber.
+        The key to authenticate the subscriber. Containing a list with the keyword unsubscribe and the email ['unsubscribe', email].
```
ix5, line 564: Maybe mention here that the key is also base64-encoded.
The stray `['unsubscribe', email]` here is also a bit confusing.
Maybe put it on a new line with "Expected syntax" or something.

ix5 (commented):
> Except the thing with the git commit message length, im not to sure what i should do there I can drop the prefix to make them shorter, but i guess the idea is to use prefixes? I can create super short unspecific commit messages which i feel would be worse. So i am a bit lost how i should solve that any ideas?

Don't worry too much about that. Those commit message conventions are just that, _conventions_. I personally use abbreviations or more terse wording and kind of "force" myself to keep below 50 chars, as I've set my editor to "warn" me when I go over.

E.g.
```
views: comments: moderation: Replace Yo response with a more descriptive text
```
I'd replace with
```
views: comments: Proper response instead of 'Yo'
```

But again, that's just stylistic pet peeves. You'll notice the usefulness of short commit messages when browsing git logs and being annoyed at the cut-off messages ending in ellipses.

---

Thanks for those tests. One inline comment, everything else looks marvelous!

isso/views/comments.py
ix5: Pushed to your branch with my suggested change. I've left the base64 encoding part out because there's also a signature attached, which would just be confusing.

ix5:
Verified using `make test` and `npm i apidoc; npx apidoc --config apidoc.json --input isso/ --output _apidoc/`

Thank you for your contribution!

isso/views/comments.py
fliiiix: thanks :+1:

PR #808, opened 2022-03-06
setup.py: drop python3.5 support

fliiiix:
Since not tests are run for 3.5 drop support
with the next release.
Also Python 3.5 has reached end-of-life on 5-Sep-2020.
Python 3.5.10 is the final release of 3.5.
https://www.python.org/downloads/release/python-3510/


fliiiix:
Technically even 3.6 is unsupported but as long as there is no need to drop support i would argue its continent to still support it.

fliiiix:
This change should probably be noted in some changelog?

CHANGES.rst
```diff
@@ -3,6 +3,9 @@ Changelog for Isso
 
 %(version)s (%(date)s)
 
+- Drop support for outdate Python version 3.5
```
ix5, line 6: outdated

ix5 (changes requested):
Please also change `docs/docs/install.rst` which mentions 3.5 as well

ix5:
Hey @fliiiix, you're also very welcome to join the `#isso` IRC channel, preferably via [Matrix: #isso](https://matrix.to/#/#isso:libera.chat) or on [Libera.Chat](https://libera.chat/) directly.

fliiiix:
I will update this PR hopefully on the weekend.

ix5:
While you're at it, maybe also test compatibility with 3.11 and advertise that?

CHANGES.rst
fliiiix: done

fliiiix:
3.11 is not release yet so i would wait for the official release

ix5:
> 3.11 is not release yet so i would wait for the official release

My thinking was that we can already put it into the GH actions test matrix before the release is out, but not yet adversise support in `setup.py`.

Doesn't matter too much though, merged.

PR #809, opened 2022-03-06
CHANGES.rst: Fix small typos

fliiiix:


fliiiix:
fair point but i guess simple fixing simple typos should always be ok, but i would agree that you don't want to change more that that

PR #810, opened 2022-03-06
Bugfix/api docs

fliiiix:
some small api docs changes


ix5:
Verified using:
```sh
npm install apidoc
npx apidoc --config apidoc.json --input isso/ --output _apidoc/
firefox _apidoc/index.html
```
(we probably ought to document this procedure somewhere and also upload those generated apiDocs somewhere)

Thanks for the contribution!

PR #811, opened 2022-03-06
__init__.py: Fix config startup check

fliiiix:
gunicorn is not using main() it uses make_app via run.py.
Because of that we should check for the config file inside
make_app.
This is probably also true for other deployment options.

### DIFF
--- a/isso/__init__.py
+++ b/isso/__init__.py
@@ -177,6 +177,10 @@ def make_app(conf=None, threading=True, multiprocessing=False, uwsgi=False):
 
     isso = App(conf)
 
+    if not any(conf.getiter("general", "host")):
+        logger.error("No website(s) configured, Isso won't work.")
+        sys.exit(1)
+
     # check HTTP server connection
     for host in conf.getiter("general", "host"):
         with http.curl('HEAD', host, '/', 5) as resp:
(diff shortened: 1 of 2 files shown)


fliiiix:
I hope there is a better solution than this

fliiiix:
Good point i will add a way to reproduces this issue. I ran into that when testing the last pre-release.

fliiiix:
Here is the example:

on master branch
```
$ gunicorn -b localhost:8080 -w 4 --preload isso.run             
2022-03-17 20:20:28,908 WARNING: unable to connect to your website, Isso will probably not work correctly. Please make sure, Isso can reach your website via HTTP(S).
[2022-03-17 20:20:28 +0100] [800014] [INFO] Starting gunicorn 20.1.0
[2022-03-17 20:20:28 +0100] [800014] [INFO] Listening at: http://127.0.0.1:8080 (800014)
[2022-03-17 20:20:28 +0100] [800014] [INFO] Using worker: sync
[2022-03-17 20:20:28 +0100] [800015] [INFO] Booting worker with pid: 800015
[2022-03-17 20:20:28 +0100] [800018] [INFO] Booting worker with pid: 800018
[2022-03-17 20:20:29 +0100] [800019] [INFO] Booting worker with pid: 800019
[2022-03-17 20:20:29 +0100] [800020] [INFO] Booting worker with pid: 800020
```

on this branch
```
gunicorn -b localhost:8080 -w 4 --preload isso.run
2022-03-17 20:21:22,747 ERROR: No website(s) configured, Isso won't work.
```

ix5:
Looks like a sensible change. Thanks!

Issue #442, opened 2018-05-12
textarea class conflicts with CSS libraries


mycahp:
Just a suggestion:

the `textarea` class conflicts with some CSS libraries. In my case, Bulma. It caused the actual textarea to not expand when content was being entered into the box.

I fixed this by adding `-isso` to all the `textarea` class references in the embed.js file. 

You might consider adopting a more specific class name like this so that it CSS libraries don't cause this type of behavior.

ix5:
Heads up: Changing existing isso CSS will break many user setups who extend or replace isso's CSS with their own.

ix5:
Tracked in # 816

PR #816, opened 2022-03-20
css, js: Prefix all directives with 'isso-'

ix5:
This "namespaces" Isso's CSS in order not to conflict with things on page

Fixes https://github.com/posativ/isso/issues/442


ix5:
Also, while I have your attention: @Kerumen @adroste

adroste:
@ix5 This is definitely a breaking change regarding custom overrides. So anything below v0.13 could lead to mad users.

ix5:
Would you mind going over the change and testing whether I missed anything or fat-fingered some string?

adroste:
Looks good (without testing it locally).

At first I thought you missed the input form selectors, but they are limited by the postbox scope.

https://github.com/posativ/isso/blob/88653210660af2069329896f8fb05d603b9881f4/isso/js/app/isso.js#L88-L91

Nevertheless, you could also prefix the keys for the items that are persisted in the localstorage:

https://github.com/posativ/isso/blob/88653210660af2069329896f8fb05d603b9881f4/isso/js/app/isso.js#L122-L124

ix5:
> Nevertheless, you could also prefix the keys for the items that are persisted in the localstorage:
> 
> https://github.com/posativ/isso/blob/88653210660af2069329896f8fb05d603b9881f4/isso/js/app/isso.js#L122-L124

That's a great catch, thank you! Moving to prefixed key names will ofc invalidate any username/email/website users have set, but imo it's not worth to write a migration for that.

(Note that I hope you just happened to review the version I pushed after I rebased it on https://github.com/posativ/isso/pull/822 ;-)

PR #817, opened 2022-03-20
docs, demo: <noscript> to warn that JS is needed

ix5:
Add `<noscript>` tags with a warning that for comments to work, Javascript needs to be enabled.

If a browser has Javascript enabled (as almost all modern browsers do by default), the tags will be invisible to the user.


ix5:
Feel free to improve upon the `<noscript>` message.

PR #822, opened 2022-03-21
[WIP] isso: js: Replace pug templates with plain js

ix5:
This removes dependencies on `pug`, `pug-loader` and `pug-runtime` and reduces the js dist bundle to ~56Kb.

This also makes it possible to test templating functionality using `Jest`.

Note: Javascript destructuring assignment is not available in IE, so this commit drops any compatibility Isso might've had.
See https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment#browser_compatibility


ix5:
A lot of my pending Jest tests rely on this, so I'll be selfish and merge this even if it's quite ugly.

isso/js/app/templates/comment-loader.js
```diff
@@ -0,0 +1,9 @@
+var html = function (globals) {
+  // De-structure globals dict
+  ({comment, pluralize} = {...globals});
+  return "" +
+"<div class='isso-comment-loader' id='isso-loader-${comment.name}'>"
++ "<a class='load_hidden' href='#'>" + pluralize('comment-hidden', comment.hidden_replies) + "</a>"
++ "</div>"
```
Kerumen, line 7: Wouldn't had be cleaner to use template string for this?

```js
return `
  <div class='isso-comment-loader' id='isso-loader-${comment.name}'>
    <a class='load_hidden' href='#'>${pluralize('comment-hidden', comment.hidden_replies)}</a>
  </div>
`
```

Kerumen, line 7: Wouldn't had be cleaner to use template string for this?

```js
return `
  <div class='isso-comment-loader' id='isso-loader-${comment.name}'>
    <a class='load_hidden' href='#'>${pluralize('comment-hidden', comment.hidden_replies)}</a>
  </div>
`
```

Kerumen: While writing this, I see that you forgot to interpolate `${comment.name}`! Be careful.

Kerumen: While writing this, I see that you forgot to interpolate `${comment.name}`! Be careful.

ix5: > Wouldn't had be cleaner to use template string for this?

I did consider using template literals, see https://github.com/ix5/isso/tree/template-literals, but I saw browser compat wasn't too great, see [MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals#browser_compatibility)

But if we're using destructuring assignments, that argument kind of goes out of the window anyway.

> While writing this, I see that you forgot to interpolate ${comment.name}! Be careful.

Yup, way ahead of you ;) https://github.com/posativ/isso/commit/727dbc333dcef71aa1b9d4571021769223290291

isso/js/app/templates/comment-loader.js
```diff
@@ -0,0 +1,9 @@
+var html = function (globals) {
+  // De-structure globals dict
+  ({comment, pluralize} = {...globals});
```
ix5, line 3: Do you know any cleaner way to extract only `comment` and `pluralize` from the `globals` object, @Kerumen ? I only settled on using destructuring because _some guy on stackoverflow_ said to use it.

Kerumen: You can simply do:

```js
var { comment, pluralize } = globals;
```

On a side note, as you are now using destructuring and perhaps template literals, you could switch `var` to `const` / `let`, no? Let's write modern JS! 😄

isso/js/app/templates/comment-loader.js
Kerumen: For your question, I would have done the same but extracting some repetitions into variables before the template string. This way the template is a bit cleaner. For example you are using 2 times ``${comment.author ? comment.author : i18n('comment-anonymous')}`. I would have done: 

```js
const author = comment.author ? comment.author : i18n('comment-anonymous');

return `
  <span class='author'>${author}</span>
`;
```

isso/js/app/templates/comment-loader.js
ix5: I can understand your enthusiasm, but I'm also wary of breaking Isso for a lot of people who are stuck on old browsers (or their assistive technologies don't understand ES6 yet, or they're on locked-down ancient corporate machines, etc.).

Is transpiling to older JS version a thing that can be relied upon? Such that we might emit an ES5-compatible `embed.legacy.js` through `babel`?

isso/js/app/templates/comment-loader.js
ix5: Doh! That's so obvious in hindsight... Thank you.

PR #825, opened 2022-03-21
treewide: Remove manpage generation

ix5:
They clutter the repo and configs unnecessarily.
Since Isso is not being packaged in debian any more, we have little reason to keep them.

Improve docs and commmand line reference instead.


HLFH:
I am packaging isso for ArchLinux though: https://aur.archlinux.org/packages/isso
But ok, I removed the manpages as well in the PKGBUILD.

Issue #292, opened 2016-12-28
Does not work on Windows (wfastcgi+IIS)


jtanx:
I was trying running this on Azure, and I ran into an issue with the use of `AF_UNIX` in [wsgi.py](https://github.com/posativ/isso/blob/master/isso/wsgi.py#L198)

Replacing it with:

```python
address_family = socket.AF_INET
```

Seems to fix it, but then I'm not sure what that's used for (or if it's even used?)

posativ:
What have you exactly tried to run Isso on Azure? I think the default web server doesn't use this wsgi class.

jtanx:
I was running Isso as a Web App following [this](https://docs.microsoft.com/en-us/azure/app-service-web/web-sites-python-configure) guide. It uses wfastcgi and IIS, It works as long as I remove the reference to `AF_UNIX` sockets.

This is my wsgi entry point:

```python
#!/usr/bin/env python
import site, os

from isso import make_app
from isso import config as isso_config

ISSO_CONFIG=os.path.abspath("production.cfg")

application = make_app(isso_config.load(ISSO_CONFIG))
```

jtanx:
Hmm, looks like there are a few other issues:

```python
Unhandled exception in wfastcgi.py: Traceback (most recent call last):
  File "D:\home\Python27\wfastcgi.py", line 833, in main
    result.close()
  File "D:\home\Python27\wfastcgi.py", line 710, in __exit__
    self.send(FCGI_END_REQUEST, zero_bytes(8), streaming=False)
  File "D:\home\Python27\wfastcgi.py", line 755, in send
    send_response(self.stream, self.record.req_id, FCGI_STDOUT, self.header_bytes)
  File "D:\home\Python27\wfastcgi.py", line 372, in send_response
    raise TypeError("content must be encoded before sending: %r" % content)
TypeError: content must be encoded before sending: u'Status: 200 Ok\r\nContent-Type: text/plain\r\nAccess-Control-Allow-Origin: --\r\nAccess-Control-Allow-Credentials: true\r\nAccess-Control-Allow-Methods: HEAD, GET, POST, PUT, DELETE\r\nAccess-Control-Allow-Headers: Origin, Referer, Content-Type\r\nAccess-Control-Expose-Headers: X-Set-Cookie, Date\r\n\r\n'
```

[...]

PR #831, opened 2022-03-27
Bugfix/gravatar display avatar size

fliiiix:
* Fix display for larger avatar images
* limit default gravatar images to the same 55px as the non gravatar images

I didn't see any negative effects of removing the 7% with but idk i guess it was there for a reason
Screenshot 2022-03-27 at 20-54-21 Binärgewitter Talk #292 USB Autobatterie
Screenshot 2022-03-27 at 20-53-51 Binärgewitter Talk #292 USB Autobatterie

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -9,6 +9,11 @@ Changelog for Isso
   This change necessitates adjusting custom CSS files to the new names.
 - Drop support for outdated Python version 3.5
   (#808, l33tname)
+- Fix avatar sizing, limit default gravatar images to 55px (#831, l33tname)
+  In case of a custom gravatar URL, the ``&s=55`` size parameter will have
+  to be added, see `Gravatar: Image requests`_.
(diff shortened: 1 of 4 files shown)


ix5:
You might want to document that `?s=x` sets the size of the generated avatar, and maybe also link to [Gravatar.com: Image requests](http://en.gravatar.com/site/implement/images/) docs.

It should be noted that the default dimensions for Gravatar seem to be `80x80` pixels, documenting that might also be a good idea.

Re: The `7%` width, I can't think of a valid reason for that either. Removing is fine.

Just one nit: `isso.css with limit of 7% leads to weird spacing` isn't a commit message, it's a statement. `css: isso.css: Remove width=7% from div.avatar` is better. Then you can explain the rationale in the message body.

fliiiix:
thanks for picking this up :tada: i don't had time to work on it

PR #835, opened 2022-03-31
Removes reference to unmaintained yaourt

Fryboyter:
The AUR helper yaourt has been unmaintained for years, has bugs and is probably not compatible with the current version of pacman. Yaourt should therefore no longer be used.

Since there are many alternative AUR helpers and an installation is also possible without such a tool, I think we should not give an installation example.


ix5:
Please write a proper commit message, e.g. `docs: install: Remove reference to yaourt`, with the message body containing the justification you posted above.

Issue #836, opened 2022-04-20
Docker won't build docker image


Gymnae:
After a fresh clone, I receive the following error:
```bash
# docker build ~/services/isso/git/. -t isso_local
Sending build context to Docker daemon  4.378MB
Step 1/16 : FROM node:dubnium-buster AS jsBuilder
 ---> c90eecc338a8
Step 2/16 : WORKDIR /src/
 ---> Using cache
 ---> 7af30b93c790
Step 3/16 : COPY . .
 ---> b6101fc6177e
Step 4/16 : RUN make init js
 ---> Running in 5b29c106970f
npm install
npm WARN read-shrinkwrap This version of npm is compatible with lockfileVersion@1, but package-lock.json was generated for lockfileVersion@2. I'll try to do my best with it!
npm ERR! code EBADPLATFORM
npm ERR! notsup Unsupported platform for fsevents@2.3.2: wanted {"os":"darwin","arch":"any"} (current: {"os":"linux","arch":"x64"})
npm ERR! notsup Valid OS:    darwin
npm ERR! notsup Valid Arch:  any
npm ERR! notsup Actual OS:   linux
npm ERR! notsup Actual Arch: x64

npm ERR! A complete log of this run can be found in:
npm ERR!     /root/.npm/_logs/2022-04-20T18_42_22_229Z-debug.log
make: *** [Makefile:39: init] Error 1
```

My host is a Linux x64 Debian 11 system

BBaoVanC:
I'm getting this same issue, also on a Debian 11 system.

ix5:
A quick [Google-Fu](https://stackoverflow.com/questions/66093284/unsupported-platform-for-fsevents2-3-2-installation-issue) reveals that the OS inside the container most likely has a version of **`npm` that is too old**.

Currently, the Isso `Dockerfile` uses Debian "Buster" as a base for js-related stuff, which seems to bundle a very old version of node and npm.

I've been meaning to revamp the `Dockerfile` for a while now.

How's this for you? Using Alpine instead of Debian/Ubuntu:
```Dockerfile
# Isso production Dockerfile

# First stage: Build Javascript client parts using NodeJS

FROM node:current-alpine AS isso-js
WORKDIR /src/

# make is not installed by default on alpine
RUN apk add --no-cache make

# Only copy necessities, to not trigger re-builds unnecessarily
COPY ["Makefile", "package.json", "package-lock.json", "./"]
COPY ["isso/js/", "./isso/js/"]
[...]

PR #841, opened 2022-04-23
Dockerfile: Use Alpine base images, optimize process

ix5:
The [current `Dockerfile`](https://github.com/posativ/isso/blob/master/Dockerfile#L2) uses an outdated version of Debian which ships versions of `node` and `npm` that incompatible with some of the packages Isso requires.

Issues reported in https://github.com/posativ/isso/issues/836

Fix this by using newer base images which ship the required node/npm versions. As a side benefit, Alpine Linux images are smaller, thus resulting in faster build times (less time required to pull) and a smaller generated docker image for Isso.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,37 +1,87 @@
-# First, compile JS stuff
-FROM node:dubnium-buster AS jsBuilder
+# Isso production Dockerfile
+
+# First stage: Build Javascript client parts using NodeJS
+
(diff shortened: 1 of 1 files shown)


ix5:
You can verify this change using `docker build -t isso .` and then running it with e.g.:
```sh
docker run -d --rm --name isso -p 8080:8080 -v $PWD/config:/config -v $PWD/db:/db isso:latest
```

PR #555, opened 2019-06-07
Fix the wsgi response

SnorlaxYum:

Issue #524, opened 2019-01-20
Fails to render Postbox if require-email = true


rocka:
I upgraded my isso instance to git master branch nowadays, only to find that the Postbox won't be rendered; there is an error in the console:

```
Uncaught TypeError: Cannot read property 'replace' of null
```

After some research, I found that bacause of #356, Postbox `<input>`s does not have `placeholder` attributes now. But following code:

https://github.com/posativ/isso/blob/4fe24296a2235de8d063cc3df7a93a54cc9d077f/isso/js/app/isso.js#L55-L64

requires the input to have a placeholder to work. We should either change this, or revert #356

jelmer:
@mpchadwick thoughts?

Laul0:
same issue ?

```
TypeError: null is not an object (evaluating 'e("[name='email']",m).getAttribute("placeholder").replace')
```

jelmer:
Rather than trying to fix this and get it wrong again, I've just reverted it. Open to new merge proposals re-swapping placeholders for labels after 0.12.2.

PR #844, opened 2022-04-24
Make it possible to identify the page author's comments

### LINKED ISSUE #82: Identifying the author of the document

BBaoVanC:
Fixes #82 

## TODO

- [x] Documentation
- [x] CHANGES.rst entry
- [x] Possibly remove the colors and instead explain how to change them in the docs
- [x] Screenshot using default Isso theme

## Summary of changes

- Page author's comments will have a different background
- Their name will have a different color
- Configured using `data-isso-page-author-hashes` field in the `<script>` tag; this can include multiple hashes.
- Add matching CSS rules, and change the margin/padding rules a bit to make `.isso-text-wrapper` take up the comment's size
- Move `.isso-follow-up` div outside of `.isso-text-wrapper`

Screenshot at the bottom of this description.

## Moving `.isso-follow-up`

I moved the `.isso-follow-up` div to be outside of `.isso-text-wrapper` (so it is instead a direct child of `.isso-comment`). Then I gave the background color to the `.isso-text-wrapper`. Without this change, if someone else replies to a comment by the page author, their comment still has the background. This is because the `.isso-text-wrapper` div would be as big as the parent comment plus the follow-up ones, and the follow-up comments have a transparent background (instead of explicitly setting one).

I couldn't think of any better solution.

Luckily this change only required [changing a couple lines in `isso.js`](https://github.com/posativ/isso/pull/844/files#diff-34a9f4a1ee05a7d28372e0370d8f38e273f1b97dc84c0b943ae9916c5f16995c) so it would properly match the new location.

## Configuring

Configure by setting the `data-isso-page-author-hashes` variable to the hash of the user (or a list of them) that should be highlighted. If providing multiple, separate them with a comma, space, or both.

For example, these all work (I made up the hashes though):

- `data-isso-page-author-hashes="86g7n8g67nm,8m787mg8"`
- `data-isso-page-author-hashes="86g7n8g67nm 8m787mg8"`
- `data-isso-page-author-hashes="86g7n8g67nm, 8m787mg8"`

This is parsed using [code from this article](https://www.tutorialspoint.com/split-space-delimited-string-and-trim-extra-commas-and-spaces-in-javascript).

## Screenshot with the theming shown in docs:

Screenshot 2022-05-07 at 14-20-11 Isso Demo

```css
.isso-comment.isso-is-page-author > .isso-text-wrapper {
    background-color: #bae0ea;
}

.isso-comment.isso-is-page-author > .isso-text-wrapper > .isso-comment-header > .isso-author {
    color: #19798d;
}
```

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -18,6 +18,8 @@ Changelog for Isso
   in post box (`#861 <https://github.com/posativ/isso/pull/861>`_, ix5)
 - Re-enable ``no-intra-emphasis`` misaka extension in default config.
 - Allow ``sup`` and ``sub`` HTML elements by default
+- Add ``data-isso-page-author-hashes`` option to client which makes it possible
+  to style comments and replies made by the page's author(s).
 
 .. _Gravatar: Image requests: http://en.gravatar.com/site/implement/images/
 
--- a/docs/docs/reference/client-config.rst
+++ b/docs/docs/reference/client-config.rst
@@ -20,6 +20,7 @@ preferably in the script tag which embeds the JS:
             data-isso-vote="true"
             data-isso-vote-levels=""
             data-isso-feed="false"
+            data-isso-page-author-hashes="f124cf6b2f01,7831fe17a8cd"
             data-isso-reply-notifications-default-enabled="false"
             src="/prefix/js/embed.js"></script>
 
@@ -142,6 +143,49 @@ data-isso-feed
     Enable or disable the addition of a link to the feed for the comment
     thread. The link will only be valid if the appropriate setting, in
     ``[rss]`` section, is also enabled server-side.
+
+data-isso-page-author-hashes
+    Provide the hash (or list of hashes) of the current page's author. Any
+    comments made by those authors will be given the ``isso-is-page-author``
+    class. This can be styled using CSS.
+
+    The hash of a user can be found by checking the ``data-hash`` parameter on the
+    ``<div>`` tag containing their comment. This is what the element looks like:
+
+    .. code-block:: html
+
+        <div class="isso-comment isso-no-votes" id="isso-14" data-hash="41faef0a49fc">
+
+    According to this example, your script tag would look something like this:
+
+    .. code-block:: html
+
+        <script src="..." data-isso-page-author-hashes="41faef0a49fc"></script>
+
+    When adding multiple hashes to support multiple page authors, separate the
+    hashes by a command and/or space. All of the following are acceptable
+    (although the hashes are made up):
+
+    - ``data-isso-page-author-hashes="86g7n8g67nm,8m787mg8"``
+    - ``data-isso-page-author-hashes="86g7n8g67nm 8m787mg8"``
+    - ``data-isso-page-author-hashes="86g7n8g67nm, 8m787mg8"``
+
+    For example, these CSS rules make the page author's name a sort of
+    turquoise color, and the comment's background a lighter version of that:
+
+    .. code-block:: css
+
+        .isso-comment.isso-is-page-author > .isso-text-wrapper {
+            background-color: #bae0ea;
+        }
+
+        .isso-comment.isso-is-page-author > .isso-text-wrapper > .isso-comment-header > .isso-author {
+            color: #19798d;
+        }
+
+
+    .. versionadded:: 0.13
+
 
 data-isso-reply-notifications-default-enabled
     Set to ``true`` to make the reply notifications checkbox on the postbox be
--- a/isso/css/isso.css
+++ b/isso/css/isso.css
@@ -92,6 +92,10 @@
     font-weight: bold;
     color: #555;
 }
+.isso-comment > .isso-text-wrapper > .isso-comment-header .isso-page-author-suffix {
+    font-weight: bold;
+    color: #2c2c2c;
+}
 .isso-comment > .isso-text-wrapper > .isso-textarea-wrapper .isso-textarea,
 .isso-comment > .isso-text-wrapper > .isso-textarea-wrapper .isso-preview {
     margin-top: 0.2em;
--- a/isso/css/isso.css
+++ b/isso/css/isso.css
@@ -33,8 +33,6 @@
 
 #isso-root .isso-comment {
     max-width: 68em;
-    padding-top: 0.95em;
-    margin: 0.95em auto;
 }
 #isso-root .isso-preview .isso-comment {
     padding-top: 0;
@@ -43,11 +41,12 @@
 #isso-root .isso-comment:not(:first-of-type),
 .isso-follow-up .isso-comment {
     border-top: 1px solid rgba(0, 0, 0, 0.1);
+    margin-bottom: 0.5em;
 }
 .isso-comment > .isso-avatar {
     display: block;
     float: left;
-    margin: 3px 15px 0 0;
+    margin: 0.95em 0.95em 0;
 }
 .isso-comment > .isso-avatar > svg {
     max-width: 48px;
@@ -58,6 +57,7 @@
 }
 .isso-comment > .isso-text-wrapper {
     display: block;
+    padding: 0.95em;
 }
 .isso-comment .isso-follow-up {
     padding-left: calc(7% + 20px);
--- a/isso/js/app/config.js
+++ b/isso/js/app/config.js
@@ -22,7 +22,8 @@ var config = {
                   "#be5168", "#f19670", "#e4bf80", "#447c69"].join(" "),
     "vote": true,
     "vote-levels": null,
-    "feed": false
+    "feed": false,
+    "page-author-hashes": "",
 };
 
 var js = document.getElementsByTagName("script");
@@ -81,5 +82,8 @@ config["langs"] = languages;
 // code outside this file should look only at langs
 delete config["lang"];
 delete config["default-lang"];
+
+// Convert page-author-hash into a array by splitting at whitespace and/or commas
+config["page-author-hashes"] = config["page-author-hashes"].split(/[\s,]+/);
 
 module.exports = config;
--- a/isso/js/app/isso.js
+++ b/isso/js/app/isso.js
@@ -151,7 +151,7 @@ var insert_loader = function(comment, lastcreated) {
         entrypoint = $("#isso-root");
         comment.name = 'null';
     } else {
-        entrypoint = $("#isso-" + comment.id + " > .isso-text-wrapper > .isso-follow-up");
+        entrypoint = $("#isso-" + comment.id + " > .isso-follow-up");
         comment.name = comment.id;
     }
     var el = $.htmlify(template.render("comment-loader", {"comment": comment}));
@@ -208,7 +208,7 @@ var insert = function(comment, scrollIntoView) {
     if (comment.parent === null) {
         entrypoint = $("#isso-root");
     } else {
-        entrypoint = $("#isso-" + comment.parent + " > .isso-text-wrapper > .isso-follow-up");
+        entrypoint = $("#isso-" + comment.parent + " > .isso-follow-up");
     }
 
     entrypoint.append(el);
--- a/isso/js/app/templates/comment.js
+++ b/isso/js/app/templates/comment.js
@@ -38,8 +38,8 @@ var html = function (globals) {
      + "<a class='isso-edit' href='#'>" + i18n('comment-edit') + "</a>"
      + "<a class='isso-delete' href='#'>" + i18n('comment-delete') + "</a>"
   + "</div>" // .isso-comment-footer
-  + "<div class='isso-follow-up'></div>"
 + "</div>" // .text-wrapper
++ "<div class='isso-follow-up'></div>"
 + "</div>" // .isso-comment
 };
 module.exports = html;
--- a/isso/js/app/templates/comment.js
+++ b/isso/js/app/templates/comment.js
@@ -7,9 +7,11 @@ var html = function (globals) {
   var svg = globals.svg;
 
   var author = comment.author ? comment.author : i18n('comment-anonymous');
+  var isPageAuthor = conf["page-author-hashes"].indexOf(comment.hash) > -1;
+  var pageAuthorClass = (isPageAuthor ? " isso-is-page-author" : '');
 
   return "" +
-"<div class='isso-comment' id='isso-" + comment.id + "'>"
+"<div class='isso-comment" + pageAuthorClass + "' id='isso-" + comment.id + "' data-hash='" + comment.hash + "'>"
 + (conf.gravatar ? "<div class='isso-avatar'><img src='" + comment.gravatar_image + "'></div>" : '')
 + (conf.avatar ? "<div class='isso-avatar'><svg data-hash='" + comment.hash + "'</svg></div>" : '')
 + "<div class='isso-text-wrapper'>"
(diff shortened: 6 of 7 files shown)


BBaoVanC:
I'm trying to think of the best way to handle it. One way could be to just make it the job of the website owner to add their own CSS that styles the element based on the `data-hash` field. Although that would pretty much limit it just to changing colors/backgrounds/etc. Ideally there should be a badge of some sort, probably something server side. Maybe a field in the client named `data-isso-page-owner` or something.

ix5:
Related: https://github.com/posativ/isso/pull/321, a previous attempt, sadly abandoned by the original author.

BBaoVanC:
@ix5 If you're still here, where should I get an author icon from? I found [this one](https://uxwing.com/writer-blogger-author-icon/) from a quick search, is that okay to use in this? (It says "License: All icons are free to use any personal and commercial projects without any attribution or credit.")

BBaoVanC:
Actually, I think just having a different name color and background color will be better.

BBaoVanC:
I think it's mostly implemented now, I just need to figure out some padding.

BBaoVanC:
I just pushed a commit that fixes that padding, but I also discovered a different issue.

The follow-up comments are inside the main comments div. That means that the follow up comments (which are transparent background) also look like they're from the author, even if they're not. I fixed this temporarily by setting an explicit background for comments that aren't by the page author, but I don't think that's a good fix.

BBaoVanC:
Ok I think I fixed that issue. I had to move the `isso-follow-up` div to be outside of the `isso-text-wrapper`, which luckily only required changing a couple lines in one of the JS files to match its new location.

BBaoVanC:
Oops, I messed up the rule so it was still highlighting the follow up comments and didn't notice. It's fixed now, see the screenshot at the top of the page in the pull request description.

BBaoVanC:
That latest commit allows you to add multiple page authors to be highlighted by putting multiple hashes separated by a comma or a space (or both). I used a solution listed on this page to parse it: https://www.tutorialspoint.com/split-space-delimited-string-and-trim-extra-commas-and-spaces-in-javascript

ix5:
Hi, thanks for your work. I'll review the merits of this later, just in the meantime, please try to keep the level of messages down a bit. You can edit your original message by clicking the three-dot button on the top right corner instead of writing a stream-of-consciousness barrage, which will make it quite hard for people to quickly decipher the core points later on.

Your commits should be squashed (or at least split into logical units) and should also mention the name of the component they're touching. See https://github.com/posativ/isso/pull/805#pullrequestreview-899518668 or https://github.com/posativ/isso/pull/831#issuecomment-1080027228

BBaoVanC:
> Hi, thanks for your work. I'll review the merits of this later, just in the meantime, please try to keep the level of messages down a bit. You can edit your original message by clicking the three-dot button on the top right corner instead of writing a stream-of-consciousness barrage, which will make it quite hard for people to quickly decipher the core points later on.
> 
> Your commits should be squashed (or at least split into logical units) and should also mention the name of the component they're touching. See [#805 (review)](https://github.com/posativ/isso/pull/805#pullrequestreview-899518668) or [#831 (comment)](https://github.com/posativ/isso/pull/831#issuecomment-1080027228)

Ok, I fixed that now, check the main description for the pull request.

Also is this commit message okay?

```
isso: Add page-author-hash option to highlight page author's comments

- Give page author's name a different color
- Give the comment a different background
- Supports multiple page authors
- Moved `.isso-follow-up` outside of `.isso-text-wrapper`
```

I feel like it would be weird to split the CSS and JS changes apart since they both require eachother for the feature to work correctly.

isso/css/isso.css
```diff
@@ -92,6 +96,9 @@
     font-weight: bold;
     color: #555;
 }
+.isso-comment > .isso-text-wrapper > .isso-comment-header .isso-page-author {
+    color: #19798d !important;
```
ix5, line 100: Please don't use `!important`, either make the selector more specific or fix the underlying issue.

isso/css/isso.css
```diff
@@ -58,10 +58,14 @@
 }
 .isso-comment > .isso-text-wrapper {
     display: block;
+    padding: 0.95em;
 }
 .isso-comment .isso-follow-up {
     padding-left: calc(7% + 20px);
 }
+.isso-comment.isso-page-author > .isso-text-wrapper {
+    background-color: #bae0ea;
```
ix5, line 67: How did you arrive at that specific color? Will it also look good on dark background?

isso/js/app/config.js
```diff
@@ -81,4 +82,6 @@ config["langs"] = languages;
 delete config["lang"];
 delete config["default-lang"];
 
+config["page-author-hash"] = config["page-author-hash"].split(/[\s,]+/);
```
ix5, line 85: Please comment in the code what this does, it's not very obvious to JS beginners at first glance
`// Split "page-author-hash" by comma and spaces into array`

isso/js/app/templates/comment.js
```diff
@@ -7,15 +7,16 @@ var html = function (globals) {
   var svg = globals.svg;
 
   var author = comment.author ? comment.author : i18n('comment-anonymous');
+  var page_author_class = (conf["page-author-hash"]).indexOf(comment.hash) > -1 ? " isso-page-author" : "";
```
ix5, line 10: `isso-is-page-author` instead? (implying that it's kind of a boolean to CSS editors)

ix5:
Re: commits and splitting:

> Ok, I fixed that now, check the main description for the pull request.
> 
> Also is this commit message okay?
> 
> ```
> isso: Add page-author-hash option to highlight page author's comments
> 
> - Give page author's name a different color
> - Give the comment a different background
> - Supports multiple page authors
> - Moved `.isso-follow-up` outside of `.isso-text-wrapper`
> ```
> 
> I feel like it would be weird to split the CSS and JS changes apart since they both require eachother for the feature to work correctly.

You can split the commits, most logical would be one commit for each component or for one coherent/atomic change, e.g. the CSS modifications. What I meant is that you need to get rid of something that looks like:
- `Initial commit`
- `Oops, fix typo`
- `Rework component X`
- `Fix test`
- `Revert 'Rework component X`

This is hard to review. Especially a lot of "Fix X" and then "Fix Y" stuff.

I hope you understand what I mean. For instance, you introduced a `is_page_author` variable in on of the commits in your [previous history](https://github.com/posativ/isso/commits/685221c8563800cc1fecd2297dec3567f5900ae0) and then removed it again. 

Just structure the commits like you structured the headlines in the (now edited and much more clear) original post.

ix5 (changes requested):
Please document the new option extensively and add a well-written `CHANGES` entry.

Would also be nice to have screenshots of the thing in action with the default colors, not your dark theme.

isso/css/isso.css
BBaoVanC: I used a color picker to get roughly the same hue as the link color (on the reply button), and then I lightened and desaturated it.

As for it working on a dark background, it doesn't really fit. However, it looked to me that the rest of the CSS doesn't really plan for a dark background by default either. The background color of the postbox fields and the preview field are set to be very light colors which doesn't fit in on a dark background. Since the text color isn't overridden inside the preview box, it becomes white text on a light background if the outer site is set to have white text on black background. Additionally, the text color when hovering over links in (such as the reply button) is very dark, making it hard to read. A lot of these colors need to be overridden to work well on a dark background.

reply text hovering on black background

white text on light background with preview

ix5: To be honest, maybe don't include any style changes in this PR then and let users decide.

For instance, I'm using this and would have to tweak my CSS to override the default styling that this PR introduces:
Screenshot from 2022-04-27 01-38-55

ix5: Also, there should be some kind of textual indication for blind/impaired users instead of a purely visual one.
Screenshot from 2022-04-27 01-41-56

The `(Author)` text should then be part of `isso/js/app/i18n/en.js` as e.g.
`"postbox-page-author": "(Author)" // or "(Site owner)"`

BBaoVanC: > To be honest, maybe don't include any style changes in this PR then and let users decide.
> 
> For instance, I'm using this and would have to tweak my CSS to override the default styling that this PR introduces: Screenshot from 2022-04-27 01-38-55

I think the commits I just pushed a few minutes ago fixes that due to a [small refactor I did](https://github.com/posativ/isso/compare/0b64ba1866d24b66b0feee2126a016cbac2969b2..a85c85ea159db5f585f6bd9d1e4b08a5e7370ab2?diff=split#diff-c20d9e95fdd0eb015d78774a734e4b2b708656bb788869a2d46140d8d7f152b1).

And I guess I can just take those colors out and instead put them somewhere in the docs anyways.

ix5: I hope you're not waiting on feedback from my end.

For now, let's leave out the background color changes (also, those might clash with the highlight-linked-comment PR?).

To gather screenshots, you can use the pre-packaged demo site. Run `isso -c share/isso-dev.cfg run` and visit `localhost:8080/demo`. Edit `isso/demo/index.html` to add the author badge `data-isso-` attribute.

BBaoVanC: > I hope you're not waiting on feedback from my end.
> 
> For now, let's leave out the background color changes (also, those might clash with the highlight-linked-comment PR?).
> 
> To gather screenshots, you can use the pre-packaged demo site. Run `isso -c share/isso-dev.cfg run` and visit `localhost:8080/demo`. Edit `isso/demo/index.html` to add the author badge `data-isso-` attribute.

Yeah I wasn't waiting for anything, I've just been putting it off because of other projects I'm working on, and school. I can probably finish this tomorrow though, if not today.

CHANGES.rst
```diff
@@ -14,6 +14,8 @@ Changelog for Isso
   to be added, see `Gravatar: Image requests`_.
 - Add ``data-isso-reply-notifications-default-enabled`` option to configure
   whether or not the subscribe to replies checkbox should be checked by default.
+- Add `data-isso-page-author-hashes` option to client which makes it possible
```
ix5, line 17: `reST` requires double backticks (\`\`data-isso-...\`\`) around code instead of markdown's one.

isso/js/app/templates/comment.js
```diff
@@ -7,16 +7,19 @@ var html = function (globals) {
   var svg = globals.svg;
 
   var author = comment.author ? comment.author : i18n('comment-anonymous');
+  var isPageAuthor = conf["page-author-hashes"].indexOf(comment.hash) > -1;
+  var pageAuthorClass = (isPageAuthor ? " isso-is-page-author" : '');
+  var pageAuthorSuffix = (isPageAuthor ? " (Author)" : '');
```
ix5, line 12: This is a string that must be translated. Add it to `i18n/en.js` (and - in a separate commit - to any other translation files of languages you speak)

~~Also, I think we might have a problem with right-to-left languages. With e.g. `Amir` + ` (Author)` in Arabic (`أمير` and `(مؤلف)`, this should be `أمير (مؤلف)`~~ Huh, Javascript apparently already does the correct thing for us and keeps the Arabic part RTL, but the latin part ltr. Nice.

isso/js/app/isso.js
```diff
@@ -151,7 +151,7 @@ var insert_loader = function(comment, lastcreated) {
         entrypoint = $("#isso-root");
         comment.name = 'null';
     } else {
-        entrypoint = $("#isso-" + comment.id + " > .isso-text-wrapper > .isso-follow-up");
+        entrypoint = $("#isso-" + comment.id + " > .isso-follow-up");
```
ix5, line 154: You might want to write a test case for this - template code and `isso.js` relying on DOM selectors such as `$('#isso-x')` has caused trouble before, e.g. https://github.com/posativ/isso/issues/524

See https://isso-comments.github.io/docs/technical-docs/testing-client/ (latest docs not uploaded to posativ.org yet) and `isso/js/tests/unit/` for examples.

ix5 (changes requested):
Thanks for tackling this again!

Just so that you don't forget: There should be an explanation how to derive the hash and what the hash means.

isso/js/app/templates/comment.js
ix5: Also, the ` (Author)` part should have a different color. Otherwise, it wouldn't stick out

And everyone could just type in `BBao (Author)` into the comment box - you'd probably catch it when manually unlocking comments, but still annoying for people without moderation turned on.

isso/js/app/isso.js
BBaoVanC: I took a look at those docs and the other files, but I'm still not really sure what I should do to test it. It doesn't look like a similar scenario to the other tests.

ix5: I'll write a test for it myself.

docs/docs/reference/client-config.rst
```diff
@@ -143,6 +144,49 @@ data-isso-feed
     thread. The link will only be valid if the appropriate setting, in
     ``[rss]`` section, is also enabled server-side.
 
+data-isso-page-author-hashes
+    Provide the hash (or list of hashes) of the current page's author. Any
+    comments made by those authors will be given the ``isso-is-page-author``
+    class. This can be styled using CSS.
+
+    The hash of a user can be found by checking the ``data-hash`` parameter on the
```
ix5, line 152: Well done, very nicely documented!

Note to myself: The specifics of this should probably be moved into the "Tips & Tricks" section. I'll do that afterwards, not part of this PR.

Also, a script to derive the hashes should be added to contrib or as a CLI option that takes `salt` and other set options into account (also by me):
```python
import codecs
from hashlib import pbkdf2_hmac as pbkdf2
val = "email@email.com".encode("utf-8")  # change me!
salt = b"Eech7co8Ohloopo9Ol6baimi"
iterations = 1000
dklen = 6
func = "sha1"
compute = pbkdf2(
    hash_name=func,
    password=val,
    salt=salt,
    iterations=iterations,
    dklen=dklen
)
hash_ = compute
uhash = codecs.encode(hash_, "hex_codec").decode("utf-8")
print("uhash: %s" % uhash)
```

isso/js/app/templates/comment.js
ix5: This is Farsi with a Farsi name ("Farid"-> "فرید"):    `فرید•نویسنده•17 دقیقه پیش`
This is Farsi with a Latin-script name ("John"):         `John•نویسنده•17 دقیقه پیش`

"Author" ("نویسنده") will confusingly be placed at the rightmost edge if the Author name is Latin-script. "نویسنده" highlighted in red for this screenshot:

Screenshot from 2022-05-07 19-24-05

Not all that important for this PR, we'd probably be better off with proper RTL support by dynamically setting `direction: rtl` for all of `#isso-thread` if detected/set language is an RTL one.

isso/css/isso.css
```diff
@@ -33,8 +33,6 @@
```
ix5, line 33: This commit is titled `css: isso.css: Implement styling for .isso-is-page-author` but it does not add any styling to `.isso-is-page-author`.

Here's what makes most sense: Style the `.isso-page-author-suffix` class:
```css
.isso-comment > .isso-text-wrapper > .isso-comment-header .isso-page-author-suffix {
    font-weight: bold;
    color: #2c2c2c;
}
```
Bold font weight to match author class and color is author color `#555` slightly darkened.

ix5 (changes requested):
Some small things still to fix, but all in all, looking good.

isso/css/isso.css
BBaoVanC: Oops, that happened because I squashed the commit where I removed the explicit color styling into that one and didn't rename it. I updated it now

docs/docs/reference/client-config.rst
BBaoVanC: Do I need to add that script or will you do it in another commit? And also is there anything else that needs to be modified in this PR?

ix5: I'll add that script later. Apart from the missing `.isso-page-author-suffix` styles I mentioned inline, I think this should be good to go.

All in all, this PR is a good one - I know that having a proper user management would ofc be the better option, but within the restrictions of the current project status, this is what makes most sense.

BBaoVanC: The `.isso-page-author-suffix` style has been added in 63d2198ea2e7fe1c70fa44f3a240d832b4865abf.

ix5:
Merged. Thank you for so diligently applying my feedback, @BBaoVanC !

Issue #842, opened 2022-04-23
[client] `fetchComments` is adding another PostBox even if there is already one displayed


Kerumen:
Everytime we call `window.Isso.fetchComments()`, a new PostBox is added, even if there is already a PostBox displayed. It means that if we call it 4 times, there will be 4 PostBox displayed.

<img width="961" alt="Screenshot 2022-04-23 at 15 43 20" src="https://user-images.githubusercontent.com/5436545/164901535-25274051-e169-45c4-a56c-4077bd63698c.png">

ix5:
Recent changes made to `isso.js` which moved the logic of the postbox creation to after the `/config` endpoint had been fetched introduced this issue.

Currently, Isso promises the `window.Isso.fetchComments()` and `window.Isso.init()` functions to users [in the docs](https://github.com/posativ/isso/blob/46f1bc851868e406a97c4ab3b9a7f024508503e7/docs/docs/extras/advanced-integration.rst#asynchronous-comments-loading).
Should we keep it that way? IMO those hooks make sense for SPA and the like, although I have to admit I have very little experience with the subject.

@Kerumen, would you like to tackle this issue? In the process, you might use the opportunity to untangle the JS a bit (there is a lot of logic inside `isso.js` that might better be in separate files, afaics, and some `require()` dependency chains seem unnecessarily deep). I defer to your knowledge and experience there.

ix5:
Also, the styling to the comment box in the screenshot you attached looks very cool! Very elegant and understated.

Maybe you could **share the CSS you used**, and **if people agree**, we could make something like it (with better contrast than grey-on-grey text, and without the cool-but-too-whimsical submit button) the **Isso default?**

ix5:
@Kerumen I noticed that I still had https://github.com/posativ/isso/pull/821 open. I changed the logic there and pushed just now, so maybe you could review that PR again, with the new changes.

Kerumen:
> Recent changes made to `embed.js` which moved the logic of the postbox creation to after the `/config` endpoint had been fetched introduced this issue.
> 
> Currently, Isso promises the `window.Isso.fetchComments()` and `window.Isso.init()` functions to users [in the docs](https://github.com/posativ/isso/blob/46f1bc851868e406a97c4ab3b9a7f024508503e7/docs/docs/extras/advanced-integration.rst#asynchronous-comments-loading). Should we keep it that way? IMO those hooks make sense for SPA and the like, although I have to admit I have very little experience with the subject.
> 
[...]

Issue #771, opened 2022-02-10
Isso always calls count even if not needed


GHSam:
I'm not using comment counts on my page but the isso embed.js code is sending an empty request for them.

It looks like [the count code is missing a check](https://github.com/posativ/isso/blob/a7c1069cf5feab721a166e7796ce725287a46189/isso/js/app/count.js#L24) to see if there are any matching URLs before making the request.

GHSam:
@ix5 just noticed you've added the needs-contributor label. I can send a PR for this if you want?

ix5:
That'd be great! Please explain your rationale though, since I'm still not quite sure what the exact behavior _should_ be.

GHSam:
I was thinking just a test to check if the urls array is empty and if it is, don't bother sending the request as it does nothing if it is empty.

Something like:

```js
// No URLs matched so don't send count request as nothing to count
if (!urls.length) {
    return;
}
```

ix5:
maybe @zaygraveyard is the right person to chime in on this

ix5:
@GHSam would you like to submit a PR with @zaygraveyard's suggestion applied?

ix5:
Tracked in #869

Issue #552, opened 2019-05-08
Log out of admin interface


locness3:
Would be great to have the possibility to sign out of the admin interface via a simple button.

ix5:
Tracked in #870

Issue #465, opened 2018-07-28
Bugs with indented code


pozitron57:
1. Indented with 4 spaces code is displayed without an indent when seen in the Edit mode.
2. In admin panel such a code is displayed in one line.
3. If the comment has been edited from admin panel, it is no more possible for commenter to remove it from the page: `Failed to load resource: the server responded with a status of 403 () embed.min.js:18 Not authorized to remove this comment!`. Not sure what the default behavior should be in this case.

I made a [gif](https://i.imgur.com/lOwVEek.gif) to illustrate these bugs. I find the most important that in the admin panel there is no line breaks, any comment displayed in one line.

pozitron57:
It seems that it can be solved by applying a css property `white-space: pre-wrap` to `div.textarea` or to `div.text div` for the admin panel.

Jenselme:
I encountered this today: one comment was not formatted properly on site and I wanted to edit it in the admin. But in the admin, no line breaks are respected in the display and worst they are lost on save.

I gave your solution a try and it fixes display, which is nice. But if I edit the comment, add line breaks and saves, the line breaks are lost. From what I understand, it comes from the fact that we rely on a div with `editableContent` set to `true`. So, if I originally have:
```html
<div id="isso-text-202">
Test 1
</div>
```
I'll get after the edit:
```html
<div id="isso-text-202" class="" contenteditable="false">
<div>Test 1</div>
<div>Test 2</div>
</div>
```

The display is correct (because by default `div` are displayed as block elements) but when we extract the text with `textContent`, we loose the line breaks that are visually conveyed by the `div`.

I worked on a fast workaround for the admin: use `textarea` instead of editable divs. By default it is disabled so the user can't edit it. If you click on *Edit* it becomes editable. It seems to fix both the display and edit issues. It's not very pretty but it works. You can see it here: https://github.com/Jenselme/isso/commit/e5a1b2ac3b6692e7f1aabb8a1a21b8f028acd28c

To fix this in the "standard edit" on the website, I think you could do something similar: when the user is editing, we rely on a `textarea`. When not (comment saved or in preview), we display the rendered HTML. We could also apply this method to the admin to view rendered comments by default and still be able to edit them correctly.

Let me know what you think. If it sounds like a good idea to you, I should be able to complete the fix for the admin and the standard edit page.

Jenselme:
Any thoughts on that?

ix5:
[...]

PR #845, opened 2022-04-24
[client] Highlight linked comments to identify them

### LINKED ISSUE #838: [client] When scrolling to a linked comment, it should have some highlight
When using a link such as `https://example.com/post-1/#isso-4` for example, the comment that is being linked to should flash a couple times to make it identifiable from other nearby comments.

BBaoVanC:
Fixes #838 

TODO:

- [x] Figure out the best way to do styling
- [x] Docs
- [x] Tests (check that `.isso-target` gets applied properly)

### DIFF
--- a/isso/css/isso.css
+++ b/isso/css/isso.css
@@ -162,6 +162,19 @@
     display: none;
 }
 
+/* "target" means the comment that's being linked to, for example:
+ * https://example.com/blog/example/#isso-15
+ */
+.isso-target {
+    animation: isso-target-fade 5s ease-out;
+}
+@keyframes isso-target-fade {
+    0% { background-color: #eee5a1; }
+    /* This color should be changed when used on a dark background,
+     * maybe #3f3c1c for example
+     */
+}
+
 .isso-postbox {
     max-width: 68em;
     margin: 0 auto 2em;
--- a/isso/js/embed.js
+++ b/isso/js/embed.js
@@ -54,6 +54,19 @@ function init() {
     // and the call to fetch those is in fetchComments()
     isso_thread.append(heading);
     isso_thread.append('<div id="isso-root"></div>');
+
+    window.addEventListener('hashchange', function() {
+        if (!window.location.hash.match("^#isso-[0-9]+$")) {
+            return;
+        }
+
+        var existingTarget = $(".isso-target");
+        if (existingTarget != null) {
+            existingTarget.classList.remove("isso-target");
+        }
+
+        $(window.location.hash + " > .isso-text-wrapper").classList.add("isso-target");
+    });
 }
 
 function fetchComments() {
@@ -105,6 +118,11 @@ function fetchComments() {
             if (window.location.hash.length > 0 &&
                 window.location.hash.match("^#isso-[0-9]+$")) {
                 $(window.location.hash).scrollIntoView();
+
+                // We can't just set the id to `#isso-target` because it's already set to `#isso-[number]`
+                // So a class `.isso-target` has to be used instead, and then we can manually remove the
+                // class from the old target comment in the `hashchange` listener.
+                $(window.location.hash + " > .isso-text-wrapper").classList.add("isso-target");
             }
         },
         function(err) {
--- a/docs/docs/guides/tips-and-tricks.rst
+++ b/docs/docs/guides/tips-and-tricks.rst
@@ -166,6 +166,21 @@ From `blog.mdosch.de <https://blog.mdosch.de/2018/05/20/isso-ip-adressen-woechen
 them to to zero.*
 
 
+Change linked comment highlight color
+-------------------------------------
+
+When you click a link that goes to a specific Isso comment (such as ``https://example.com/example-post/#isso-43``),
+the comment will be highlighted yellow and then fade out over a few seconds.
+
+Add the following CSS to your site to change that color (replace ``#3f3c1c`` with any color of your choice):
+
+.. code-block:: css
+
+   @keyframes isso-target-fade {
+     0% { background-color: #3f3c1c; }
+   }
+
+
 .. attention::
 
    This section of the Isso documentation is incomplete. Please help by expanding it.
--- a/isso/js/tests/integration/highlight-comments.test.js
+++ b/isso/js/tests/integration/highlight-comments.test.js
@@ -1,0 +1,45 @@
+const ISSO_ENDPOINT = process.env.ISSO_ENDPOINT ?
+  process.env.ISSO_ENDPOINT : 'http://localhost:8080';
+
+beforeEach(async () => {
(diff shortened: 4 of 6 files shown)


BBaoVanC:
I meant to make this a WIP pull request, but forgot to before submitting.

BBaoVanC:
@ix5 Can you review this please?

I have CSS added to set a background color included. I don't think that separating it out like what we did in #844 makes sense because this feature should be included by default. It's also different since it doesn't require any configuration by the site owner to function.

Also, I did figure out how to get it not to conflict with #844. By not setting the 100% keyframe, it automatically just fades out to whatever background color is the fallback for the element.

Let me know if there's a better place for the `hashchange` event listener, and if `window.addEventListener` is the right function to use for it; I wasn't really sure what to do with it.

ix5:
Hey @BBaoVanC, I'm off to a well-deserved break, starting tomorrow. So I won't be able to review your changes for about a week.

Let's do a deal: You figure out a solution and submit draft PRs to at least some of these open issues which I think you're more than capable of solving:
- https://github.com/posativ/isso/issues/111
- https://github.com/posativ/isso/issues/448
- https://github.com/posativ/isso/issues/465
- https://github.com/posativ/isso/issues/859
- Race conditions / re-arranging parts in https://github.com/posativ/isso/pull/821 and https://github.com/posativ/isso/issues/842

... and I'll be happy to take some time to review this highlighting PR once I get back ;-)

Would also be nice if at some of these low hanging fruits could be fixed:
- https://github.com/posativ/isso/issues/552
- https://github.com/posativ/isso/issues/771

BBaoVanC:
> Hey @BBaoVanC, I'm off to a well-deserved break, starting tomorrow. So I won't be able to review your changes for about a week.
> 
> Let's do a deal: You figure out a solution and submit draft PRs to at least some of these open issues which I think you're more than capable of solving:
> 
>     * [Make name/URL/e-mail editable #111](https://github.com/posativ/isso/issues/111)
> 
>     * [Counter problem #448](https://github.com/posativ/isso/issues/448)
> 
>     * [Bugs with indented code #465](https://github.com/posativ/isso/issues/465)
> 
>     * [[client] Clean up and standardize client CSS #859](https://github.com/posativ/isso/issues/859)
> 
>     * Race conditions / re-arranging parts in [[client] js: Fetch public conf once from /config; embed.js: Preserve default config values #821](https://github.com/posativ/isso/pull/821) and [[client] `fetchComments` is adding another PostBox even if there is already one displayed #842](https://github.com/posativ/isso/issues/842)
> 
> 
> ... and I'll be happy to take some time to review this highlighting PR once I get back ;-)
> 
> Would also be nice if at some of these low hanging fruits could be fixed:
> 
>     * [Log out of admin interface #552](https://github.com/posativ/isso/issues/552)
> 
>     * [Isso always calls count even if not needed #771](https://github.com/posativ/isso/issues/771)

I'll take a look at those. Hope you have a nice break!

isso/css/isso.css
```diff
@@ -159,6 +159,13 @@
     display: none;
 }
 
+.isso-target {
+    animation: target-fade 5s ease-out;
+}
+@keyframes target-fade {
```
ix5, line 165: That animation name should be isso-namespaced

isso/css/isso.css
```diff
@@ -159,6 +159,13 @@
...
+    0% { background-color: #eee5a1; }
```
ix5, line 166: This color (light yellow) is quite jarring on dark background. Try e.g. the demo with `body { background: black; color: white; }`

Also, maybe add a little code comment such as `Flash comment light yellow when focussed`, either here or in embed.js

isso/js/embed.js
```diff
@@ -105,6 +112,9 @@ function fetchComments() {
             if (window.location.hash.length > 0 &&
                 window.location.hash.match("^#isso-[0-9]+$")) {
                 $(window.location.hash).scrollIntoView();
```
ix5, line 114: Wrap this in a try-catch block (ES5 has it, I think?) or check if `$(window.location.hash` actually returns an element. `fetchComments` (and `init`) is a function that definitely should not throw.

isso/js/embed.js
```diff
@@ -105,6 +112,9 @@ function fetchComments() {
...
+
+                // We can't just set the id because that would overwrite the existing `isso-X` id
```
ix5, line 116: Could you elaborate on what you meant by that code comment?

ix5 (commented):
> I have CSS added to set a background color included. I don't think that separating it out like what we did in #844 makes sense because this feature should be included by default. It's also different since it doesn't require any configuration by the site owner to function.

That's fine. As commented inline, the yellow is a bit jarring at night, but yellow is universally accepted as a marker/highlight/attention color. Maybe once there's a dark mode someone can style it differently. Keep as-is for now.

> Let me know if there's a better place for the hashchange event listener, and if window.addEventListener is the right function to use for it; I wasn't really sure what to do with it.

I'm not qualified to comment on event listeners, my JS knowledge is pretty bad. I found [this SO answer](https://stackoverflow.com/questions/6390341/how-to-detect-if-url-has-changed-after-hash-in-javascript) which seems to suggest a similar thing. Availability of [Caniuse: hashchange](https://caniuse.com/hashchange) seems good.

As for stuffing this piece of code into `init()`: IMO that file is already quite overloaded. Maybe shoving everything related to timers and events into a separate file (that can be stubbed/mocked) would simplify testing a good bit.

Several places rely on `window.location.*`. Unifying that would be a nice first step.
E.g. this is just a stupid duplication:
https://github.com/posativ/isso/blob/fdcadaddcd87fe97fbd7ca328a2ef7ea56eccb87/isso/js/app/api.js#L7
https://github.com/posativ/isso/blob/fdcadaddcd87fe97fbd7ca328a2ef7ea56eccb87/isso/js/embed.js#L74

isso/js/embed.js
BBaoVanC: Yep, can't believe I didn't catch this error when I was testing. (It happens if you click a comment for the first time, before another one was linked, meaning there is no .isso-target)

ix5:
Would you mind adding a small styling guide to the docs? (Perhaps under "Tips & Tricks", 1-2 lines should be enough).

Code-wise, this looks good right now.

BBaoVanC:
> Would you mind adding a small styling guide to the docs? (Perhaps under "Tips & Tricks", 1-2 lines should be enough).
> 
> Code-wise, this looks good right now.

I added some docs, now all that's left is a small test if possible which just makes sure that `.isso-target` gets applied when going to like `#isso-1` for example.

ix5:
Added some tests. Thank you and looking forward to reviewing other great contributions!

PR #846, opened 2022-04-25
[client] js: Support enabling reply notifications checkbox by default

### LINKED ISSUE #837: [client] Option to enable email notifications by default
I would like to make the checkbox to get email notifications enabled by default instead of needing to manually click it.

BBaoVanC:
Fixes #837 

Configure by setting `data-isso-reply-notifications-default-enabled` to `"true"` in the client settings. If set to true, then the checkbox for enabling reply notifications will be checked by default:

the checkbox is checked

### DIFF
--- a/isso/js/app/templates/postbox.js
+++ b/isso/js/app/templates/postbox.js
@@ -1,8 +1,10 @@
 var html = function (globals) {
   var i18n = globals.i18n;
+  var conf = globals.conf;
   var author = globals.author;
   var email = globals.email;
   var website = globals.website;
+  var notify = conf["reply-notifications-default-enabled"] ? " checked" : '';
 
   return "" +
 "<div class='isso-postbox'>"
@@ -39,7 +41,7 @@ var html = function (globals) {
   + "</section>"
   + "<section class='isso-notification-section'>"
     + "<label>"
-      + "<input type='checkbox' name='notification' />" + i18n('postbox-notification')
+      + "<input type='checkbox'" + notify + " name='notification' />" + i18n('postbox-notification')
     + "</label>"
   + "</section>"
 + "</div>"
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -12,6 +12,8 @@ Changelog for Isso
 - Fix avatar sizing, limit default gravatar images to 55px (#831, l33tname)
   In case of a custom gravatar URL, the ``&s=55`` size parameter will have
   to be added, see `Gravatar: Image requests`_.
+- Add ``data-isso-reply-notifications-default-enabled`` option to configure
+  whether or not the subscribe to replies checkbox should be checked by default.
 
 .. _Gravatar: Image requests: http://en.gravatar.com/site/implement/images/
 
--- a/docs/docs/reference/client-config.rst
+++ b/docs/docs/reference/client-config.rst
@@ -20,6 +20,7 @@ preferably in the script tag which embeds the JS:
             data-isso-vote="true"
             data-isso-vote-levels=""
             data-isso-feed="false"
+            data-isso-reply-notifications-default-enabled="false"
             src="/prefix/js/embed.js"></script>
 
 Furthermore you can override the automatic title detection inside
@@ -141,6 +142,18 @@ data-isso-feed
     Enable or disable the addition of a link to the feed for the comment
     thread. The link will only be valid if the appropriate setting, in
     ``[rss]`` section, is also enabled server-side.
+
+data-isso-reply-notifications-default-enabled
+    Set to ``true`` to make the reply notifications checkbox on the postbox be
+    checked by default. Otherwise, the user will have to manually opt-in to
+    reply notifications.
+
+    This setting will have no effect if ``reply-notifications`` are not enabled
+    on the server.
+
+    Default: ``false``
+
+    .. versionadded:: 0.13
 
 
 Deprecated Client Settings
--- a/isso/js/app/config.js
+++ b/isso/js/app/config.js
@@ -11,6 +11,7 @@ var config = {
     "require-email": false,
     "require-author": false,
     "reply-notifications": false,
+    "reply-notifications-default-enabled": false,
     "max-comments-top": "inf",
     "max-comments-nested": 5,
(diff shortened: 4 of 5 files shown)


isso/js/app/isso.js
```diff
@@ -71,6 +71,9 @@ var Postbox = function(parent) {
     var email_edit = function() {
         if (config["reply-notifications"] && $("[name='email']", el).value.length > 0) {
             $(".isso-notification-section", el).show();
+            if (config["reply-notifications-default-enabled"]) {
```
ix5, line 74: This should be implemented in `isso/js/app/templates/postbox.js`

Something like:

```diff
diff --git a/isso/js/app/templates/postbox.js b/isso/js/app/templates/postbox.js
index a59d6a1..3f0a94b 100644
--- a/isso/js/app/templates/postbox.js
+++ b/isso/js/app/templates/postbox.js
@@ -1,8 +1,10 @@
 var html = function (globals) {
   var i18n = globals.i18n;
+  var conf = globals.conf;
   var author = globals.author;
   var email = globals.email;
   var website = globals.website;
+  var notify = conf['reply-notifications-default-enabled'] ? ' checked' : '';
 
   return "" +
 "<div class='isso-postbox'>"
@@ -39,7 +41,7 @@ var html = function (globals) {
   + "</section>"
   + "<section class='isso-notification-section'>"
     + "<label>"
-      + "<input type='checkbox' name='notification' />" + i18n('postbox-notification')
+      + "<input type='checkbox" + notify + " name='notification' />" + i18n('postbox-notification')
     + "</label>"
   + "</section>"
 + "</div>"
```

ix5 (changes requested):
Nice work! A few changes are needed in order for this to be accepted.

Please document the change in the docs under `docs/docs/configuration/client.rst`.
My suggestion: `Set to "true" to tick the checkbox for reply notifications in the postbox by default. Per default, the checkbox will not be ticked. N.b.: You still need to enable reply notifications on the server side.`

Please add `CHANGES` entry

A few tests would be nice, but not required

BBaoVanC:
> A few tests would be nice, but not required

How would I write a test for that? Unless I missed one, I don't see any tests that test client side features, so I don't really have anything to go off of.

docs/docs/configuration/client.rst
```diff
@@ -232,6 +232,16 @@ data-isso-reply-notifications
 
 Set to `true` when reply notifications is configured with `reply-notifications = true`.
 
+data-isso-reply-notifications-default-enabled
+---------------------------------------------
+
+Set to `true` to make the reply notifications checkbox on the postbox be checked
```
ix5, line 238: Beautifully written, much more clear than my suggestion. :+1:

ix5:
> How would I write a test for that? Unless I missed one, I don't see any tests that test client side features, so I don't really have anything to go off of.

Isso uses `Jest` for testing. Sadly, there's no end-to-end testing yet, but you can write unit tests which run inside a simulated browser-like environemnt (using `jsdom`).

Here's a suggestion:
```javascript
// isso/js/tests/unit/postbox.test.js
"use strict";

// Set up our document body
document.body.innerHTML =
  '<div id=isso-thread></div>' +
  '<script ' +
    'src="http://isso.api/js/embed.min.js" ' +
    'data-isso="/" ' +
    'data-isso-reply-notifications-default-enabled="false"' +
    '></script>';

const isso = require("app/isso");
const $ = require("app/dom");
const config = require("app/config");
const i18n = require("app/i18n");
const svg = require("app/svg");
const template = require("app/template");
template.set("conf", config);
template.set("i18n", i18n.translate);
template.set("pluralize", i18n.pluralize);
template.set("svg", svg);

test('Create Postbox with reply notifications disabled by default', () => {
  var isso_thread = $('#isso-thread');
  isso_thread.append('<div id="isso-root"></div>');
  isso_thread.append(new isso.Postbox(null));

  // Will create a `.snap` file in `./__snapshots__/`.
  // Don't forget to check in those files when changing anything!
  expect(isso_thread.innerHTML).toMatchSnapshot();
});

test('Create Postbox with reply notifications enabled by default', () => {
  var script_tag = document.getElementsByTagName('script')[0];
  script_tag.setAttribute('data-isso-reply-notifications-default-enabled', 'true');

  var isso_thread = $('#isso-thread');
  isso_thread.append('<div id="isso-root"></div>');
  isso_thread.append(new isso.Postbox(null));

  // Will create a `.snap` file in `./__snapshots__/`.
  // Don't forget to check in those files when changing anything!
  expect(isso_thread.innerHTML).toMatchSnapshot();
});
```

Run `npm install` and then `npm run test-unit -- -u` to save the generated `Jest` snapshots (they'll be in `isso/js/tests/unit/__snapshots__/postbox.test.js.snap`).

isso/js/app/templates/postbox.js
```diff
@@ -1,8 +1,10 @@
 var html = function (globals) {
   var i18n = globals.i18n;
+  var conf = globals.conf;
   var author = globals.author;
   var email = globals.email;
   var website = globals.website;
+  var notify = conf["reply-notifications-default-enabled"] ? " checked" : '';
```
ix5, line 7: Maybe someone has a nicer suggestion than this ternary. But not a big deal.

ix5 (approved):
Tested locally with the setting a) not set (using default) b) set to true and c) set to false.

Each works as expected. Can't find any obvious flaws, but will leave open a few days for others to comment.

isso/js/app/templates/postbox.js
BBaoVanC: Would it be good to change it to be a boolean (just `conf["reply-notifications-default-enabled"]`), and then put the ternary inside with the other template code?

ix5: Eh, it's nitpicking and I'm not even qualified to comment on any Javascript styles. I was hoping for someone to have a better idea, but it's not important. Leave it as-is.

ix5:
Thank you for your PR, @BBaoVanC! I fixed a few things, rebased (docs moved in the meantime) and added a test.

Your other PRs have pending TODOs, if you fix them I'd be happy to merge.

PR #861, opened 2022-05-05
[client] js: Postbox: Use labels rather than placeholders

ix5:
Placeholders are is meant to be a hint (an example input) for the user about what to enter into the field, but shouldn't be used in lieu of a `<label>`.

See following links for further justification:
- https://www.tpgi.com/html5-accessibility-chops-the-placeholder-attribute/
- https://www.nngroup.com/articles/form-design-placeholders/
- https://www.maxability.co.in/2016/01/03/placeholder-attribute-and-why-it-is-not-accessible/

This commit is a reland of https://github.com/posativ/isso/pull/356 with a fix for the placeholder-replacement code in `isso.js` as reported in https://github.com/posativ/isso/issues/524

---

Also update snapshot and add a new test to avoid similar future errors.

---

**Before:**
Screenshot from 2022-05-05 15-04-33

**After:** (note `Name`, `E-Mail`, `Website` labels above input fields)

Screenshot from 2022-05-07 20-27-08

**Mobile:**
Screenshot from 2022-05-07 20-27-22

---

Previous revision (note `Preview`/`Submit` buttons vertically not aligned)
Screenshot from 2022-05-05 15-03-41

---


ix5:
Future idea: Amend the email label to also inform users that the email collected will not be published and only be used for notifications/spam avoidance (but be saved in the Isso database).
Visual:
Screenshot from 2022-05-05 15-05-38

PR #867, opened 2022-05-07
README: Include more information, new screenshot

ix5:
Add more sections, with more links, small install guide, add new screenshot.

This should make the README more inviting to people browsing GH - and also PyPI, as the README is also automatically uploaded there.

Live preview @ [TestPyPI](https://test.pypi.org/project/isso-testpypi/)

Screenshot used:
screenshot

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,14 +4,59 @@ Isso – *Ich schrei sonst* – is a lightweight commenting server written in
 Python and JavaScript. It aims to be a drop-in replacement for
 [Disqus](http://disqus.com).
 
-See **[posativ.org/isso](http://posativ.org/isso/)** for more details and
-documentation.
+## Features
 
-![Isso in Action](http://posativ.org/~tmp/isso-sample.png)
+- **Comments written in Markdown**  
+  Users can edit or delete own comments (within 15 minutes by default).
+  Comments in moderation queue are not publicly visible before activation.
+- **SQLite backend**  
+  *Because comments are not Big Data.*
+- **Disqus & WordPress Import**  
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -4,14 +4,59 @@ Isso – *Ich schrei sonst* – is a lightweight commenting server written in
...
+details and [documentation](https://posativ.org/isso/docs/).
+
+## Screenshot
+
+![Isso in Action](https://user-images.githubusercontent.com/10212877/167268553-3f30b448-25ff-4850-afef-df2f2e599c93.png)
+
+## Geting started
+
+**Requirements**
+- Python 3.6+ (+ devel headers)
```
jelmer, line 29: Why do we need devel headers or a C compiler? @ix5

ix5: This was a copy-paste from the docs. Probably necessary for misaka and CFFI things?

ix5: Devel headers were added way back in d1a0b3f6, I just happen to show up in the blame logs because I bumped versions a couple of times and moved files ;)

As for the validity, maybe we ought to collect empirical data by removing devel headers from our respective systems and seeing whether a full installation in a clean virtualenv (including misaka CFFI compilation) still works.

README.md
```diff
@@ -4,14 +4,59 @@ Isso – *Ich schrei sonst* – is a lightweight commenting server written in
...
+- SQLite 3.3.8 or later
```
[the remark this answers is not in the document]
ix5: Same as for the devel headers, let's verify and then I'm not at all opposed to removing those docs lines.

PR #869, opened 2022-05-07
js: count: Don't send `/count` request if there are no counters

### LINKED ISSUE #771: Isso always calls count even if not needed

BBaoVanC:
Fixes #771

Tasks:

- [ ] CHANGES.rst
- [ ] Docs? maybe none are needed
- [ ] Investigate returning a 400 when /count is called without any URLs
- [ ] Docs probably needs to have `POST /count` documented, that would be another PR
- [ ] Tests

### DIFF
--- a/isso/js/app/count.js
+++ b/isso/js/app/count.js
@@ -24,16 +24,18 @@ module.exports = function () {
 
     var urls = Object.keys(objs);
 
-    api.count(urls).then(function(rv) {
-        for (var key in objs) {
-            if (objs.hasOwnProperty(key)) {
+    if (urls.length > 0) {
+        api.count(urls).then(function(rv) {
+            for (var key in objs) {
+                if (objs.hasOwnProperty(key)) {
 
-                var index = urls.indexOf(key);
+                    var index = urls.indexOf(key);
 
-                for (var i = 0; i < objs[key].length; i++) {
-                    objs[key][i].textContent = i18n.pluralize("num-comments", rv[index]);
+                    for (var i = 0; i < objs[key].length; i++) {
+                        objs[key][i].textContent = i18n.pluralize("num-comments", rv[index]);
+                    }
                 }
             }
-        }
-    });
+        });
+    }
 };


ix5:
Hi, I don't think small change merits a CHANGES entry or any docs changes.

As for the other TODOs, returning a 400 for empty URL lists makes sense.
See the `POST` endpoint for relevant code section: https://github.com/posativ/isso/blob/668882f6c551c0cb979139ad62b2a97a0c968914/isso/views/comments.py#L982-L989

While at it, we can probably remove the `/count` `GET` endpoint - it's unused and not part of the `apiDoc`-generated API docs anyway (yet still part of the obsolete `rst` docs, see [docs: Server API](https://github.com/posativ/isso/blob/668882f6c551c0cb979139ad62b2a97a0c968914/docs/docs/reference/server-api.rst#get-comment-count) - you have a TODO for overhauling that as well)
https://github.com/posativ/isso/blob/668882f6c551c0cb979139ad62b2a97a0c968914/isso/views/comments.py#L960-L968

As for tests, I have tried getting my hands dirty with puppeteer-based screenshots and comparisons to snapshot png files via imagemagick `compare`.
E.g. `compare -metric AE -fuzz 5% -highlight-color White -lowlight-color Black isso-thread-prev.png isso-thread-new.png x:` or simply by comparing image signatures without metadata via `identify -quiet -format "%#" {image1,image2}.png`

ix5:
As for removing `/count` GET, see https://github.com/posativ/isso/pull/883

As for returning a 400 if thread not found: IMO there isn't currently a very smart behavior present anyway, I'd rather the endpoint for `/count` return a dict of `thread id -> count` mappings. That way, the caller doesn't have to keep state of order and the server can gracefully skip over nonexistent URLs/threads.

PR #870, opened 2022-05-08
admin: Add log out button

### LINKED ISSUE #552: Log out of admin interface

BBaoVanC:
Fixes #552

Adds a logout button to the header, in the top right.

Tasks:

- [x] CHANGES.rst

image

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -13,7 +13,9 @@ New Features
 - Add CSS variables for better organization and flexibility (`#1001`_, pkvach)
 - Add support for comment search by Thread URL in admin interface (`#1020`_, pkvach)
 - Add sorting option for comments (`#1005`_, pkvach)
+- admin: Add log out button (`#870`_, bbaovanc)
 
+.. _#870: https://github.com/posativ/isso/pull/870
 .. _#966: https://github.com/posativ/isso/pull/966
 .. _#998: https://github.com/isso-comments/isso/pull/998
 .. _#1000: https://github.com/isso-comments/isso/pull/1000
--- a/isso/css/admin.css
+++ b/isso/css/admin.css
@@ -51,6 +51,9 @@ input {
     color: #4d4c4c;
     text-decoration: none;
 }
+.header .logout {
+    float: right;
+}
 .outer {
     background-color: #eeeeee;
     box-shadow: 0 0 0.5em #c0c0c0 inset;
--- a/isso/js/admin.js
+++ b/isso/js/admin.js
@@ -124,6 +124,11 @@ function send_edit(com_id, hash, isso_host_script) {
     edit(com_id, hash, author, email, website, comment, isso_host_script);
     stop_edit(com_id, true);
 }
+function log_out() {
+    // Delete cookie
+    document.cookie = "admin-session=; Max-Age=0; domain=" + window.location.hostname + "; path=/";
+    window.location.reload();
+}
 
 function toggleTooltip(tooltipContainer) {
     const tooltipText = tooltipContainer.querySelector(".search-tooltip-text");
--- a/isso/templates/admin.html
+++ b/isso/templates/admin.html
@@ -18,6 +18,7 @@
           </a>
         </div>
       </header>
+      <a id="logout" class="logout label" onClick="javascript:log_out()">Log Out</a>
     </div>
     <div class="outer">
       <div class="filters">


isso/views/comments.py
```diff
@@ -1141,6 +1142,12 @@ def login(self, env, req):
             isso_host_script = self.isso.conf.get("server", "public-endpoint") or local.host
             return render_template('login.html', isso_host_script=isso_host_script)
 
+    def logout(self, env, req):
```
ix5, line 1145: Since we're not doing anything to invalidate the cookies (or rather, the tokens), there's no reason to ping the server for anything. Just delete the cookie on the client side (or introduce a mechanism to invalidate the cookie on the server as well).

isso/views/comments.py
```diff
@@ -1141,6 +1142,12 @@ def login(self, env, req):
...
+        response = redirect("/admin")
+        response.headers.add("Set-Cookie", "admin-session=none; path=/; max-age=0;")
+        response.headers.add("X-Set-Cookie", "isso-admin-session=none; path=/; max-age=0;")
```
ix5, line 1148: I can't remember why initially this whole `X-Set-Cookie`/`Set-Cookie` duality started...
Probably something related to proxies stripping certain headers? Or are we just cargo-culting/copy-pasting this everywhere for no good reason?

ix5 (changes requested):
See inline comments

Edit: My idea, not complete:
```diff
diff --git a/isso/css/admin.css b/isso/css/admin.css
index 3f694c3..fd2bbb5 100644
--- a/isso/css/admin.css
+++ b/isso/css/admin.css
@@ -30,16 +30,25 @@ input {
 }
 .header header {
     display: block;
-    float: left;
     font-weight: normal;
     margin-right: 16.0363%;
-    width: 41.9818%;
 }
 .header header .logo {
-    float: left;
     max-height: 60px;
     padding-right: 12px;
 }
+.header header .title {
+    display: inline;
+}
+.header header .title a {
+    display: inline-block;
+    width: 41.9818%;
+}
+.header header .logout {
+    float: right;
+    margin-top: 0.8em;
+    line-height: 1.6em;
+}
 .header header h1 {
     font-size: 1.55em;
     margin-bottom: 0.3em;
diff --git a/isso/js/admin.js b/isso/js/admin.js
index 5b1e508..8d48167 100644
--- a/isso/js/admin.js
+++ b/isso/js/admin.js
@@ -124,4 +124,11 @@ function send_edit(com_id, hash, isso_host_script) {
     edit(com_id, hash, author, email, website, comment, isso_host_script);
     stop_edit(com_id, true);
 }
-
+function log_out() {
+    // Delete cookie
+    var cookie_name = "admin-session";
+    var domain = window.location.hostname;
+    document.cookie = cookie_name + "=; Max-Age=0; domain=" + domain + "; path=/";
+    // Redirect back to /admin login page
+    window.location.href = "admin";
+}
diff --git a/isso/templates/admin.html b/isso/templates/admin.html
index 3332c5b..10dd6f9 100644
--- a/isso/templates/admin.html
+++ b/isso/templates/admin.html
@@ -16,6 +16,7 @@
             <h2>Administration</h2>
           </a>
         </div>
+        <a id="logout" class="logout label" onClick="javascript:log_out()">Log out</a>
       </header>
     </div>
     <div class="outer">
```

isso/views/comments.py
ix5: Ah, to get around cross-origin restrictions, see https://github.com/posativ/isso/commit/6691810316ef1fc445251c1a1f3e02dbcdc54df6

BBaoVanC:
Wow, the admin CSS is a horrible mess right now. is it better to just do as little change here (like I already have) so it works, and then make another pull request to clean up the CSS?

ix5:
> Wow, the admin CSS is a horrible mess right now. is it better to just do as little change here (like I already have) so it works, and then make another pull request to clean up the CSS?

Re: Admin CSS: Yes, please open another PR for that and keep this PR focused.

Re: Clearing the cookies: I'm not quite sure whether we need to match `SameSite` and `Secure` attributes from server when _clearing_ cookies. Also, IMO we can use a bit more "modern" JS and modern Interfaces such as [CookieStore](https://developer.mozilla.org/en-US/docs/Web/API/CookieStore) in the admin (just keep the client JS as backward-compatible as possible).

ix5:
@BBaoVanC would you mind applying the PR feedback? I know it's a bit much to ask, but we should at least try to keep track of SameSite and Secure stuff, I think.

jelmer:
Please rebase this PR after the force-push.

jelmer (changes requested):
Please rebase

BBaoVanC:
> Please rebase

Done

BBaoVanC:
Is there anything else this PR needs? I believe when I last looked at it I had already applied @ix5's changes.

jelmer (approved):

PR #872, opened 2022-05-09
[client] Fix avatar sizing on Chrome & Safari

BBaoVanC:
Before this change, on Chrome and Safari (but not Firefox), the `svg` inside `.isso-avatar` has a size of 0 by 0 pixels. Not sure why it worked on Firefox, but this makes sense because svg's have no native size, and since there was no minimum size specified, it defaulted to 0.

Instead of just setting `width` to `48px`, I set it to `100%` and left `max-width` set to `48px`. [It should be better practice this way](https://css-tricks.com/tale-width-max-width/).

I am wondering if setting `height` is a good idea, since it seems to have no effect. I kept it in just because it's clearer.

## Before (demo on Chromium)

the avatar image is seemingly blank as it has a size of zero pixels

## After

the avatar is displayed properly, with a size of 48x48 px

### DIFF
--- a/isso/css/isso.css
+++ b/isso/css/isso.css
@@ -51,6 +51,8 @@
 .isso-comment > .isso-avatar > svg {
     max-width: 48px;
     max-height: 48px;
+    width: 100%;
+    height: 100%;
     border: 1px solid rgba(0, 0, 0, 0.2);
     border-radius: 3px;
     box-shadow: 0 1px 2px rgba(0, 0, 0, 0.1);


ix5:
As for tests, I have tried getting my hands dirty with puppeteer-based screenshots and comparisons to snapshot png files via imagemagick `compare`.
E.g. `compare -metric AE -fuzz 5% -highlight-color White -lowlight-color Black isso-thread-prev.png isso-thread-new.png x:` or simply by comparing image signatures without metadata via `identify -quiet -format "%#" {image1,image2}.png`

ae5960e8-a6fc-491f-b252-898ecf59af95:
I'm experiencing this issue on firefox/chrome on Linux with the latest isso running on Docker.

PR #753, opened 2021-12-24
[ISSUE TEMPLATES] Add Issue Form for bug reporting, feature requests

ix5:
GitHub has recently introduced their new "Issue Forms" as a public beta. This feature allows collecting user input in a structured form and guides them towards generating actionable bugs.

For more information, please see the [public announcement](https://github.blog/changelog/2021-06-23-issues-forms-beta-for-public-repositories/) and the [documentation](https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/syntax-for-issue-forms).

For anyone wanting to test out how the form would present itself, please see https://github.com/ix5/isso/issues/new/choose

PR #873, opened 2022-05-15
.github: Add issue, PR templates

ix5:
Encourage users to fill out some basic information and guide their thought processes.

As opposed to https://github.com/posativ/isso/pull/753, these templates are much less intrusive/inquisitive and merely guide the user instead of demanding excessive amounts of information.

Reasoning: While reviewing PRs of more or less new contributors to the project such as @BBaoVanC and @fliiiix, I had to coax their PRs metadata into a more reviewable shape. BBaoVanC even ended up writing checklists for themself. Formalizing this process a bit makes sense, IMO.

@jelmer thoughts?

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -1,0 +1,21 @@
+<!-- Just like NASA going to the moon, it's always good to have a checklist when
+creating changes.
+The following items are listed to help you create a great Pull Request: -->
+## Checklist
+- [ ] All new and existing **tests are passing**
+- [ ] (If adding features:) I have added tests to cover my changes
+- [ ] (If docs changes needed:) I have updated the **documentation** accordingly.
+- [ ] I have added an entry to `CHANGES.rst` because this is a user-facing change or an important bugfix
+- [ ] I have written **proper commit message(s)**
+      <!-- Title ideally under 50 characters (at most 72), good explanations,
+      affected part of Isso mentioned in title, e.g. "docs: css: Increase font-size on buttons"
+      Further info: https://github.com/joelparkerhenderson/git-commit-message -->
+
+## What changes does this Pull Request introduce?
+<!-- Explain what this PR will do, how one can try out the changes -->
+...
+
+## Why is this necessary?
+<!-- Link to existing bugs, post reproducible setups that demonstrate the
+     issue/change -->
+...
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -1,0 +1,13 @@
+<!-- The following items are listed to help you get your issue resolved quickly -->
+## Checklist
+- [ ] I am running the latest version. Installing Isso from GitHub from the `master` branch does not fix my issue
+- [ ] I have checked the [troubleshooting guide](https://posativ.org/isso/docs/troubleshooting/)
+- [ ] I have searched the [open issues](https://github.com/posativ/isso/issues), but my issue has not already been reported
+
+## What is not working?
+<!-- Explain your issue in detail -->
+...
+
+## How can one reproduce this issue?
+<!-- Post your isso.cfg, other relevant configuration/setup -->
+...


.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,25 @@
+<!-- Just like NASA going to the moon, it's always good to have a checklist when
+creating changes.
+The following items are listed to help you create a great Pull Request: -->
+## Checklist
+- [ ] All new and existing **tests are passing**
```
jelmer, line 5: Why does this need a checklist - doesn't running the github actions cover this?

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,25 @@
...
+  - [ ] Python tests using `make test flakes` are passing
+  - [ ] Javascript unit tests using `npm run test-unit` are passing
+  - [ ] Javascript integration tests using `npm run test-integration` are passing
+- [ ] I have added tests to cover my changes
+- [ ] My change requires a change to the **documentation**.
+  - [ ] I have updated the documentation accordingly.
+  - [ ] I have added an entry to `CHANGES` because this is a user-facing change or an important bugfix
+- [ ] I have written **proper commit message(s)**
+      <!-- Title under 50 characters, good explanations, affected part of Isso
+      mentioned in title, e.g. "isso: css: Increase font-size on buttons"
```
jelmer, line 15: Why "isso: "? isn't that implied for all changes to this repo?

jelmer:
As mentioned earlier, I'm not hard opposed but also not convinced templates solve concrete problems we have.

.github/PULL_REQUEST_TEMPLATE.md
BBaoVanC: I agree, it should just be `css: Some change` or `js: Some client change` for example.

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,25 @@
...
+- [ ] All new and existing **tests are passing**
+  - [ ] Python tests using `make test flakes` are passing
+  - [ ] Javascript unit tests using `npm run test-unit` are passing
+  - [ ] Javascript integration tests using `npm run test-integration` are passing
+- [ ] I have added tests to cover my changes
+- [ ] My change requires a change to the **documentation**.
+  - [ ] I have updated the documentation accordingly.
+  - [ ] I have added an entry to `CHANGES` because this is a user-facing change or an important bugfix
+- [ ] I have written **proper commit message(s)**
+      <!-- Title under 50 characters, good explanations, affected part of Isso
```
BBaoVanC, line 14: Should 50 be the limit? Maybe 72 would be better since sometimes it's hard to fit everything under 50 characters. Many commits already on the repo are already about 55-60 characters long.

Maybe this could be reworded to "ideally under 50 characters, but no more than 72"?

.github/PULL_REQUEST_TEMPLATE.md
ix5: - Ideally, GH Actions should not be the first instance PR authors come across the tests. Getting the tests to pass by pushing to (re-)trigger CI runs would be very tedious - we'd want authors to have the tests running locally so they can understand and fix errors faster.
  I'd also wager a lot of authors most likely won't even register the tiny red/green check marks.
- Spelling the need for working tests out explicitly (even if it may be a repetition of the "contributing" section of the docs) is a good reminder for authors, even already when thinking about submitting drafts

.github/PULL_REQUEST_TEMPLATE.md
ix5: I was thinking of the client css (`isso: css: ...`) as opposed to the docs css (`docs: css`).
But I agree with both of you and have pushed with a better example.

.github/PULL_REQUEST_TEMPLATE.md
ix5: Thanks, fixed.

ix5:
I'm still waiting for an answer from posativ regarding new uploads of the docs so I can link the contributing and testing guides.

.github/PULL_REQUEST_TEMPLATE.md
jelmer: Using some sort of CI system (all of which are reported with check masks) is common for everything on GitHub. The more text you include here, the less likely people are to actually read all of it.

ix5: I've reduced the amount of text and bullet points even further.

I am convinced and have seen evidence of this type of checklist being useful - and though I hate this type of argument, it's also something other successful projects do.

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,21 @@
+<!-- Just like NASA going to the moon, it's always good to have a checklist when
+creating changes.
+The following items are listed to help you create a great Pull Request: -->
+## Checklist
+- [ ] All new and existing **tests are passing**
+- [ ] (If adding features:) I have added tests to cover my changes
+- [ ] (If docs changes needed:) I have updated the **documentation** accordingly.
+- [ ] I have added an entry to `CHANGES.rst` because this is a user-facing change or an important bugfix
+- [ ] I have written **proper commit message(s)**
```
jelmer, line 9: This is the wrong place for this, it belongs in a CONTRIBUTING.md or something similar; by the time the contributor creates the pull request they've already written these commit messages so this is the wrong moment to find out about the right format.

jelmer:
I'm still not convinced this helps, but also don't want to stand in the way so I'll abstain.

.github/PULL_REQUEST_TEMPLATE.md
ix5: All perfectly valid points. Since docs on posativ.org are outdated for now, maybe we could link to the source doc as rendered here: [isso@master:docs/contributing/index.rst](https://github.com/posativ/isso/blob/master/docs/docs/contributing/index.rst#code-contributions)

ix5:
We'll revisit once updated docs are uploaded in the proper place and we can have a `CONTRIBUTING` doc that doesn't just render [as a literal link text](https://github.com/posativ/isso/blob/878398413b91ccbbb66429506d1d1ffc5840c3d6/CONTRIBUTING.rst). For now, I'll merge this.

jelmer:
@ix after this has been live for 4 years, my experience is that it does more harm than good. 

In practice people delete the template, include it but ignore it or add filler (e.g. just adding "Closes #..." under "Why is this necessary?"). Either way I still consistently end up chasing people about following what's in the template. 

Can we either remove this or just trim it down to a reference to instructions?

PR #885, opened 2022-05-25
[server] db: comments: Constrain parent of new comments 

### LINKED ISSUE #765: [server] 500 Error with wrong parent
Hello.
I use latest stable version with Docker setup.
I found that internal sqlite table does not use any constraints on `parent` column of table comments which may be a problem in two cases.
1. DB is corrupted and one of the parents is not valid. 
2. Someone trying to add comment with invalid parent.

I tried second case with curl and got 500 Error on client and stacktrace below on server
```
2022-01-29 22:57:29,824 ERROR: POST /new
  File "/isso/lib/python3.8/site-packages/isso-0.12.5-py3.8.egg/isso/__init__.py", line 147, in dispatch
  File "/isso/lib/python3.8/site-packages/isso-0.12.5-py3.8.egg/isso/views/comments.py", line 88, in dec
    return func(self, env, req, *args, **kwargs)
  File "/isso/lib/python3.8/site-packages/isso-0.12.5-py3.8.egg/isso/views/__init__.py", line 45, in dec
  File "/isso/lib/python3.8/site-packages/isso-0.12.5-py3.8.egg/isso/views/comments.py", line 307, in new
    rv = self.comments.add(uri, data)
  File "/isso/lib/python3.8/site-packages/isso-0.12.5-py3.8.egg/isso/db/comments.py", line 57, in add
    if ref.get("parent") is not None:
AttributeError: 'NoneType' object has no attribute 'get'
```

**Steps to reproduce:**
1. Start Docker instance of isso (on http://127.0.0.1:8080/ for example)
2.  POST request http://127.0.0.1:8080/new?uri=%2Fbase.html with valid comment and wrong parent. I used "Copy as cURL" button on Firefox Web Tools and replaced `null` with some invalid id (`42` for example)

ix5:
## Checklist
      

- db: comments: Constrain parent of new comments
- tests: Add test for invalid comment parents 

## Why is this necessary?

Fixes #765

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -21,6 +21,7 @@ Changelog for Isso
 - Move ``isso-dev.cfg`` to ``contrib/`` folder
 - Run automated screenshot comparisons for testing (`#889`_)
 - wsgi: Return HTTP errors as JSON if client prefers it (`#488`_, sundbry)
+- Verify that parent of new comment is in same thread (`#885`_)
 - Add ``data-isso-page-author-hashes`` option to client which makes it possible
   to style comments and replies made by the page's author(s).
 - Add Ukrainian localisation (`#878`_, okawo80085)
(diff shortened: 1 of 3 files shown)


ix5:
Maybe this could be solved more elegantly in the future, but looks fine for now.

PR #887, opened 2022-05-26
js, templates: Replace `contenteditable` `div` with `textarea`

### LINKED ISSUE #465: Bugs with indented code

BBaoVanC:
## Checklist
- [ ] ~~(If docs changes needed:) I have updated the **documentation** accordingly.~~ N/A?
      

- Replace the `contenteditable` `div` with a proper `textarea` - fix indentation bugs

## Why is this necessary?

Fixes #465


ix5:
Hi @BBaoVanC, very nice effort, thank you! This is sorely needed.

A couple thoughts of mine:

We shouldn't add any external dependencies, instead vendor them in (the ES5-compatible version). We don't need e.g. (afaics!)
- `amd`/`module` sniffing
- the lifecycle events like `autosize:resized`, `autosize:update` etc.
- things like the `destroy` function
- the environment sniffing (nodejs/IE8, we don't support running in either, and use JSDOM for running DOM-based tests anyway)

We should set the max length HTML5 attribute for textarea (and also email and website)
https://github.com/posativ/isso/blob/46c141c9517665620929192b05d2f41275bbaefd/isso/views/comments.py#L166-L179
(And those error message responses are just plain bad - improve them and their handling on the client side as well, in another PR)

The [autosize FAQ](http://www.jacklmoore.com/autosize/#faq-slow) suggests disabling blurs on Safari for textarea - good idea to add to the CSS.

BBaoVanC:
Thanks for the response @ix5

How would I vendor the dependencies like you said? I tried to search the repository for the word `vendor` but I don't see anything.

Also about the commit messages, I'll rebase it later once it's in less of a draft stage. And in this stage with all the changes I'm trying to do it's easier to just write down the changes I make in the PR description as I go instead of writing it on every commit.

ix5:
> How would I vendor the dependencies like you said? I tried to search the repository for the word `vendor` but I don't see anything.

Just put the file into `app/lib/autosize.js`.

> Also about the commit messages, I'll rebase it later once it's in less of a draft stage. And in this stage with all the changes I'm trying to do it's easier to just write down the changes I make in the PR description as I go instead of writing it on every commit.

Dw, just something I found useful for myself.

BBaoVanC:
I decided that it's best to try and replace the "Edit" box with a postbox in a different PR. It would require a large refactor and I don't want to block this change for that long. As far as I can tell it works right now.

I removed the `editorify()` test since that function isn't needed anymore, but for some reason the screenshots don't match the CI. Could also benefit from a couple tests, which I'll look into tomorrow, in addition to rebasing and adding to `CHANGES.rst`.

Also, I feel like this project could benefit from a lot of the client code being refactored and/or rewritten. I wonder if it would be worth using TypeScript? I'll play around with this a bit in a different branch on my fork.

isso/css/isso.css
```diff
@@ -24,11 +24,8 @@
     vertical-align: bottom;
 }
 #isso-thread .isso-textarea {
-    min-height: 58px;
-    outline: 0;
-}
-#isso-thread .isso-textarea.isso-placeholder {
-    color: #757575;
+    width: 100%;
+    resize: none;
```
ix5, line 28: The `outline: 0` should stay, no? Not that I'm opposed to changing that, but any styling changes should go into another PR.

isso/js/app/isso.js
```diff
@@ -125,13 +102,12 @@ var Postbox = function(parent) {
...
         api.create($("#isso-thread").getAttribute("data-isso-id"), {
             author: author, email: email, website: website,
-            text: utils.text($(".isso-textarea", el).innerHTML),
+            text: $(".isso-textarea", el).value,
             parent: parent || null,
             title: $("#isso-thread").getAttribute("data-title") || null,
             notification: $("[name=notification]", el).checked() ? 1 : 0,
         }).then(function(comment) {
-            $(".isso-textarea", el).innerHTML = "";
-            $(".isso-textarea", el).blur();
```
ix5, line 134: Was the focus loss (`Node.blur()` taken out intentionally? I'm guessing because we don't need to re-add the placeholder text on `blur` event, no sense in triggering it?

isso/js/app/isso.js
```diff
@@ -296,16 +270,20 @@ var insert = function(comment, scrollIntoView) {
...
+                textarea.value = rv.text;
                 textarea.focus();
 
                 text.classList.remove("isso-text");
                 text.classList.add("isso-textarea-wrapper");
 
                 text.textContent = "";
                 text.append(textarea);
+
+                textarea.scrollIntoView(); // manually scroll the full textbox into view on initial resize
```
ix5, line 286: this is not "on initial resize" but rather when returning from preview to editing again, no?

isso/js/app/templates/postbox.js
```diff
@@ -10,7 +10,7 @@ var html = function (globals) {
 "<div class='isso-postbox'>"
  + "<div class='isso-form-wrapper'>"
    + "<div class='isso-textarea-wrapper'>"
-    + "<div class='isso-textarea isso-placeholder' contenteditable='true'>" + i18n('postbox-text') + "</div>"
+    + "<textarea class='isso-textarea' rows='10' maxlength='65535' placeholder='" + i18n('postbox-text') + "'></textarea>"
```
ix5, line 13: a) I can understand why you set **rows to 10** here - since we don't have auto-resizing (yet), you want the textarea to be big enough for people to enter medium-length text. But that makes the area quite long; especially on mobile already half a page length is taken up only by the postbox and it's not immediately clear that there will be actual comments when scrolling further down. Let's stick with an initial **5 rows** for now.

b) Why not add a `minlength` as well? let the browser do the validation instead of us having to do that. Should also have the benefit of localized error/hint messages instead of us having to carry translations.

isso/js/tests/unit/isso.test.js
```diff
@@ -1,60 +0,0 @@
...
-/*
- * Test goals:
- * - Test editorify()
- * - Test insert()
- * - Test insert_loader()
- * - Test Postbox()
- * Also, untangle Postbox functions from DOM element
- */
-
-test('Editorify text area', () => {
```
ix5, line 21: Please instead `test.skip()` this for now. We'll want to test the whole widget at some point in the future again.

ix5 (changes requested):
> I decided that it's best to try and replace the "Edit" box with a postbox in a different PR. It would require a large refactor and I don't want to block this change for that long. As far as I can tell, editing indentation works fine in its current state.

That's sensible, I agree.

> I removed the `editorify()` test since that function isn't needed anymore, but for some reason the screenshots don't match the CI. Could also benefit from a couple tests, which I'll look into tomorrow, in addition to rebasing and adding to `CHANGES.rst`.

Yeah, especially the `comment.png` screenshot shouldn't change, it appears it changes height by 1px. Weird. Maybe some outline sizing gets transferred.

> Also, I feel like this project could benefit from a lot of the client code being refactored and/or rewritten. I wonder if it would be worth using TypeScript? I'll play around with this a bit in a different branch on my fork.

If we decide to move to Typescript, we'd have to do a whole lot of other refactoring as well. I feel like the current state doesn't merit introducing types - it'd be like smearing lipstick on a pig. We'd also have a whole rat's tail of development dependencies to contend with. See https://github.com/posativ/isso/pull/834 for some more of my thoughts on the topic. I'm not opposed, I just feel like it should be worth it. E.g. remark42's client code is extremely complicated and inaccessible (to me).

For refactoring the code, have a look at https://github.com/isso-comments/client-experimental - maybe we can collaborate a bit there, while keeping ES5 compatibility.

**Edit:** Please also see https://github.com/posativ/isso/pull/890, I hope this makes the whole screenshot testing handling a bit easier to understand. Should I change anything to make it clearer/friendlier? Still figuring CI stuff out.

isso/js/app/isso.js
BBaoVanC: Yeah, I'm using the `placeholder` attribute on the textarea. And I feel like it would already trigger the event anyways, since clicking the submit button would make the textarea lose focus. Regardless I can't tell any difference when adding/removing that line.

isso/js/app/isso.js
BBaoVanC: This function triggers once when you press the Edit link on an existing comment. I originally added this line when I was using `autosize` to resize the textbox, and when it did that, then the first resize of the textarea would make it scroll off screen if the comment was longer than a few lines.

It actually looks like this serves a different function now. When editing without autosize, it still is halfway off-screen when editing I'll update the comment, let me know what you think about this addition.

isso/css/isso.css
BBaoVanC: I'll add it back, but I can't really tell the difference. But you're right, it can be looked at later in a CSS cleanup.

BBaoVanC:
> If we decide to move to Typescript, we'd have to do a whole lot of other refactoring as well.

I feel like this could be beneficial overall, since I'm a big fan of statically typed languages (it feels less awkward and overall more stable/definite to work in). Code is easier to follow with defined types.

It looks like with TypeScript + Webpack, we would be able to write with modern syntax, but have it transpile to ES5 compatible output. We also don't have to completely refactor everything at once since, as far as I can tell regular JS code is completely valid in TS (just you don't get the compile-time error-checking that you do with TS).

I'll look into your experimental client in a bit, but first: do we have some way to run the tests on the main Isso repo over here on the experimental client? I feel like that would be a nice way to make sure during development that the experimental client functions the same. Hopefully we could gradually add more tests, and if there's any unintended functionality on the main client which is discovered, it can be fixed.

> We'd also have a whole rat's tail of development dependencies to contend with.

I'm not too concerned about having dependencies, I feel that NPM is just sort of designed in a way that makes it hard to avoid having tons of packages/dependencies. As long as the ones we explicitly list in `package.json` are few, I think it should be ok. But I'll have to see.

---

On this PR: tests seem to work except the integration one failed with a weird error:

```
docker: Error response from daemon: cannot join network of a non running container: 04b9d43d9b947d15a78937f235c5307d37b8d1f02a08225259d083b0cbcdd0f0.
```

But it worked fine on my own machine (I used similar steps to the E2E workflow on my system to update the screenshots). I did rebase this PR onto the latest master before updating screenshots again, but also I don't really see any changes in this PR that would break that.

I'm about to rebase it again to make the commit messages better, which will trigger the CI again, so we can see if it's a random one-off error.

ix5:
The PR as it currently stands is done, as far as I can see. I know you'll fix the other issues we've talked about in a later PR.
In the interest of this not growing into a never-ending back-and-forth, I'll merge as-is (unless you have objections). I (force-)pushed a rebase and a small nit (upon leaving preview, focus textarea).

Re: Typescript etc: I'll open another issue for that so we can continue discussion there.

BBaoVanC:
> The PR as it currently stands is done, as far as I can see. I know you'll fix the other issues we've talked about in a later PR. In the interest of this not growing into a never-ending back-and-forth, I'll merge as-is (unless you have objections). I (force-)pushed a rebase and a small nit (upon leaving preview, focus textarea).
> 
> Re: Typescript etc: I'll open another issue for that so we can continue discussion there.

Sounds good to me, I don't see anything else that needs to be changed either.

ix5:
New issue opened https://github.com/posativ/isso/issues/894

> I feel like this could be beneficial overall, since I'm a big fan of statically typed languages (it feels less awkward and overall more stable/definite to work in). Code is easier to follow with defined types.
>
> It looks like with TypeScript + Webpack, we would be able to write with modern syntax, but have it transpile to ES5 compatible output. We also don't have to completely refactor everything at once since, as far as I can tell regular JS code is completely valid in TS (just you don't get the compile-time error-checking that you do with TS).

TS sure looks like a nice addition, but we might also be limiting ourselves in the contributors we get - they'd have to know TS as well. This kind of feels a bit like the "sprinkle some async" fad in python to me. Maybe start with better test coverage and work from there?

> I'll look into your experimental client in a bit, but first: do we have some way to run the tests on the main Isso repo over here on the experimental client? I feel like that would be a nice way to make sure during development that the experimental client functions the same. Hopefully we could gradually add more tests, and if there's any unintended functionality on the main client which is discovered, it can be fixed.

You can simply swap out the JS part for the integration tests (copy the generated `embed.dev.js` from the experimental client). Or do as suggested in the `README` and `cd dist && python -m http.server 8000` and change the embed URL to `localhost:8000/embed.dev.js`:
```dif
diff --git a/isso/demo/index.html b/isso/demo/index.html
index 0b043e0..b731dcc 100644
--- a/isso/demo/index.html
+++ b/isso/demo/index.html
@@ -9,8 +9,8 @@
   <div id="wrapper" style="max-width: 900px; margin-left: auto; margin-right: auto;">
    <h2><a href="index.html">Isso Demo</a></h2>
 
-   <script src="../js/embed.dev.js"
-           data-isso="../"
+   <script src="http://localhost:8000/embed.dev.js"
+           data-isso="http://localhost:8080/"
            ></script>
    <!-- Uncomment to only test count functionality: -->
    <!-- <script src="../js/count.dev.js"></script> -->
```

Then run `npm run test-integration` from the "old" isso repo.

The repo also has its own unit tests (since too much has changed). `npm run test-unit` should work from the new repo.

> I'm not too concerned about having dependencies, I feel that NPM is just sort of designed in a way that makes it hard to avoid having tons of packages/dependencies. As long as the ones we explicitly list in package.json are few, I think it should be ok. But I'll have to see.

I'm already kind of annoyed at all the Javascript packages constantly jumping sub-revisions (e.g. `Jest` and `puppeteer` all jumped several since I added them, and they need to be kept in sync with system chromium version if you don't use the `.local-chromium` version). You're never quite sure if something just broke because of a version change of a dep when re-installing or you broke something yourself. `webpack` is already quite a moving target and has not been pleasant to set up. I'm not really looking forward to spending maintainer time accepting PRs bumping revisions from bots every few days (new tsc release here, new babel release there, ...). It's not too much to do, but death by a thousand cuts.

isso/js/app/isso.js
```diff
@@ -296,9 +276,12 @@ var insert = function(comment, scrollIntoView) {
 
             toggler.canceled = false;
             api.view(comment.id, 1).then(function(rv) {
-                var textarea = editorify($.new("div.isso-textarea"));
+                var textarea = $.new("textarea.isso-textarea");
+                textarea.setAttribute("rows", 10);
```
ix5, line 280: rows still 10 here, please set to 5

BBaoVanC: Forgot about that, fixed now

ix5 (approved):

PR #903, opened 2022-06-06
migrate: Handle single newlines in WordPress comments as line breaks

projectgus:
WordPress renders a single newline in a comment as a <br> tag, but Isso renders a single newline in the comment as a single newline in the HTML. This is rendered the same as if it was a space, all text on one line.

To fix, detect single newlines when importing WordPress comments and convert to a line break in Markdown. Add a test for this also.

Example, this WordPress comment (as shown in CDATA of XML export):

> First line of comment.
> Second line of comment.

Renders in WordPress as:

> First line of comment.<br>Second line of comment.

But renders in Isso after import as if it was:

> First line of comment. Second line of comment.

After this commit is applied and comments re-imported, it renders as:

> First line of comment.
> Second line of comment.

## Checklist
- (N/A) (If docs changes needed:) I have updated the **documentation** accordingly.
- (N/A but please tell me if you disagree) I have added an entry to `CHANGES.rst` because this is a user-facing change or an important bugfix

### DIFF
--- a/isso/migrate.py
+++ b/isso/migrate.py
@@ -225,9 +225,18 @@ class WordPress(object):
         progress.finish("{0} threads, {1} comments".format(
             len(items) - skip, self.count))
 
+    def _process_comment_content(self, text):
+        # WordPress comment text renders a single newline between two blocks of
+        # text as a <br> tag, so add an explicit Markdown line break on import
+        # (Otherwise multiple blocks of text separated by single newlines are
+        # all shown as one long line.)
+        text = re.sub(r'(?!^\n)\n(?!^\n)', '  \n', text, 0)
+
+        return strip(text)
+
     def Comment(self, el):
         return {
-            "text": strip(el.find(self.ns + "comment_content").text),
+            "text": self._process_comment_content(el.find(self.ns + "comment_content").text),
             "author": strip(el.find(self.ns + "comment_author").text),
             "email": strip(el.find(self.ns + "comment_author_email").text),
             "website": strip(el.find(self.ns + "comment_author_url").text),
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -17,8 +17,10 @@ Breaking Changes
 Bugfixes & Improvements
 ^^^^^^^^^^^^^^^^^^^^^^^
 
-- TBD
+- When importing from WordPress single newlines are now converted to line breaks
+  (`#903`_, projectgus)
 
+.. _#903: https://github.com/posativ/isso/pull/903
 
 0.13.0.beta1 (2022-06-05)
 -------------------------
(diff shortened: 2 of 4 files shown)


projectgus:
I haven't touched any javascript in the PR so I don't think the js integration test failures are related. I don't have access to retry the test run, though.

ix5:
Hi @projectgus, thank you for your contribution! Especially adding a new test case is really helpful.
The e2e tests seem to have been flaky, sorry for that. Re-running (and testing the same locally), I cannot reproduce what caused that issue.

Re: Inserting a `<br>` element: Since some people disallow raw HTML in comments, I'd rather you do this the "native" Markdown way: Inserting two spaces at the end of the line, which will also render as a line break.

projectgus:
@ix5 That's funny, original version of this patch replaced with `\n\n` but then I second-guessed myself and changed to `<br>`. It's `\n\n` again now!

Also fixed the linebreak, nice catch. I think this was copypasta from the WordPress XML export.

Thanks for all your work on Isso, it's very useful. :)

isso/migrate.py
```diff
@@ -225,9 +225,18 @@ def migrate(self):
         progress.finish("{0} threads, {1} comments".format(
             len(items) - skip, self.count))
 
+    def _process_comment_content(self, text):
+        # WordPress comment text renders a single newline between two blocks of
+        # text as a <br> tag, so convert to double newlines when importing
+        # (Otherwise multiple blocks of text separated by single newlines are
+        # all shown as one long line.)
+        text = re.sub(r'(?!^\n)\n(?!^\n)', '\n\n', text, 0)
```
ix5, line 233:
```suggestion
        text = re.sub(r'(?!^\n)\n(?!^\n)', '  \n', text, 0)
```

ix5: I meant literally that you should insert two spaces at the end of the line. E.g. this:
```markdown
This is a text with 2 trailing spaces  
Next line

Two newlines should not be affected.
```
will render as
```html
<p>This is a text with 2 trailing spaces<br>
Next Line</p>
<p>Two newlines should not be affected</p>
```

projectgus: @ix5 oh snap, sorry I totally misread your first suggestion. And I clearly need to read the Markdown spec more often!

Updated, thanks for re-explaining.

ix5:
Looks good, thanks for addressing the review comments.

Please also add a `CHANGES` entry, under "bugfixes" for the "next" version (i.e. _not_ under 1.13.0.beta1), and if you like, add yourself to `CONTRIBUTORS`

projectgus:
@ix5 Done, and rebased. Thanks!

ix5:
Merged. Thank you for your contribution and don't hesitate to submit further PRs!

PR #904, opened 2022-06-12
Add trailing slash to login endpoint in login form

BBaoVanC:
## Checklist
- [ ] ~~All new and existing **tests are passing**~~
- [ ] ~~(If adding features:) I have added tests to cover my changes~~
- [ ] ~~(If docs changes needed:) I have updated the **documentation** accordingly.~~
      

Adds trailing slash to login URL in the admin login form.

## Why is this necessary?

For some reason this causes an issue logging in when running locally, but the issue doesn't happen on my live Isso instance

### DIFF
--- a/isso/templates/login.html
+++ b/isso/templates/login.html
@@ -20,7 +20,7 @@
     <main>
       <div id="login">
         Administration secured by password:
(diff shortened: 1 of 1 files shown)


ix5:
Sorry, I already pushed that.

PR #927, opened 2022-09-16
js: i18n: fix translation fallback to English

rocka:
## Checklist
      

Fix error when certain translation not present.

## Why is this necessary?

When borwser's language is not English, and certain `msgid` dose not exist in target translation, we want provide the string in English, or `[?${msgid}]` format. But current implementation would just throw an error and interrupt the whole script, because `en` is not defined.


BBaoVanC (approved):
Didn't test it myself but this looks good

ix5:
Tested an LGTM

Issue #931, opened 2022-10-18
Responsive Layout (too much width on mobile devices)


mhellmeier:
## Checklist
- [x] I am running the latest version. Installing Isso from GitHub from the `master` branch does not fix my issue
- [x] I have checked the [troubleshooting guide](https://isso-comments.de/docs/guides/troubleshooting/)
- [x] I have searched the [open issues](https://github.com/posativ/isso/issues), but my issue has not already been reported

## What is not working?

Isso isn't fully responsive. When viewing comments on a mobile device, the width is too large and content cannot fully displayed.

## How can one reproduce this issue?

See, for example, https://return2.net/laravel-auth-custom-login-validation/ on your mobile device or in a browser using the developer tools. Most comments aren't displayed correctly at the bottom of the page.

ix5:
Your site looks fine to me even with width reduced to 300px on Firefox responsive view.

Could you add a screenshot of the exact comment you're referring to? Or is your issue that the code elements your commenters pasted are not rendering as code blocks?

mhellmeier:
I just made a screenshot in the Google Chrome developer tools with the responsive view set to _iPhone 12 Pro_:


![Example](https://user-images.githubusercontent.com/1036284/197415371-6b37f25f-1ab8-4d1f-ada5-015eca41adac.png)

The form and long comments are too big for the small screen size.

ix5:
It's your own custom css causing this (weirdly only on chrome):
```javascript
// Workaround isso issue: https://github.com/posativ/isso/issues/815
    $( "<style>#isso-thread { display: table; width: 100%; } #isso-root { display: table-footer-group; }</style>" ).appendTo( "head" );
```
Remove this, https://github.com/posativ/isso/issues/815 is resolved.

mhellmeier:
Thanks for the hint but it doesn't solve the problem because really long words (for example this/very/very/very/very/very/very/very/very/very/very/very/very/long/path/on/my/personal/computer/) will crash the max width.

I will solve it in a small pull request!

PR #932, opened 2022-10-24
[client] css: add auto word break

mhellmeier:
## Checklist
      

Automatically break really long words in the text part of a comment.

I further changed the file format of the `CONTRIBUTORS` file because it is written in markdown but stored in text format.

## Why is this necessary?

Because it solves #931


ix5:
Thank you for your PR. Looks good so far.

I wonder whether the Isso renderer (`misaka`) might emit other elements that are not wrapped inside `<p>` tags, such as unordered/ordered lists etc. Those would not be caught by your CSS rule.
Could you check please?

mhellmeier:
You are totally right! Just extended the word break to the more parent class to include elements outside the `<p>` tags.

Ready for review again!

ix5:
For posterity's sake:
https://developer.mozilla.org/en-US/docs/Web/CSS/word-break
https://caniuse.com/?search=word-break
https://developer.mozilla.org/en-US/docs/Web/CSS/overflow-wrap

> Note: While `word-break: break-word` is deprecated, it has the same effect, when specified, as `word-break: normal` and `overflow-wrap: anywhere` — regardless of the actual value of the `overflow-wrap` property.

I don't think this matters too much though.

Issue #951, opened 2023-04-17
Unable to post comments with umlaut domains in the website field


schneidr:
## Checklist
- [x] I am running the latest version. Installing Isso from GitHub from the `master` branch does not fix my issue
- [x] I have checked the [troubleshooting guide](https://isso-comments.de/docs/guides/troubleshooting/)
- [x] I have searched the [open issues](https://github.com/posativ/isso/issues), but my issue has not already been reported

* **Note:** I was unable to get the master branch running, but I compared the source code and the relevant parts didn't change

## What is not working?
Posting a comment with an umlaut in the website (IDN) fails. The developer console of the browser shows the error from the server:

```
<!DOCTYPE HTML PUBLIC "-//W3C//DTD HTML 3.2 Final//EN">
<title>400 Bad Request</title>
<h1>Bad Request</h1>
<p>Website not Django-conform</p>
```

I tracked it down to the [regex definition in `views/comments.py`](https://github.com/posativ/isso/blob/64374a8af978f1cfd4a91875c18a309122c7d9ca/isso/views/comments.py#L35), which only allows `[A-Z0-9-]`.

Using the validator module instead [looks promising](https://stackoverflow.com/a/45011112/212107), otherwise the domain probably needs to be converted to IDN notation before the check.

## How can one reproduce this issue?
Enter an IDN domain with an umlaut in any installation, it also fails in the demo in the isso website, and check the developer console.

ix5:
Should be resolved by https://github.com/posativ/isso/pull/952
Otherwise, please comment if this should still be an ongoing issue.

PR #952, opened 2023-04-18
Allow umlaut domains for website addresses

schneidr:
## Checklist
      

Changed website validation to allow domain names containing umlauts

## Why is this necessary?
Resolves issue #951

### DIFF
--- a/isso/views/comments.py
+++ b/isso/views/comments.py
@@ -6,7 +6,6 @@ import time
 import functools
 import json  # json.dumps to put URL in <script>
 import pkg_resources
-import validators
 
 from configparser import NoOptionError
 from datetime import datetime, timedelta
@@ -32,12 +31,21 @@ from isso.utils.hash import md5, sha1
 from isso.views import requires
 
 
+# from Django apparently, looks good to me *duck*
+__url_re = re.compile(
+    r'^'
+    r'(https?://)?'
+    # domain...
+    r'(?:(?:[\w](?:[\w-]{0,61}[\w])?\.)+(?:[\w]{2,6}\.?|[\w-]{2,}\.?)|'
+    r'localhost|'  # localhost...
+    r'\d{1,3}\.\d{1,3}\.\d{1,3}\.\d{1,3})'  # ...or ip
+    r'(?::\d+)?'  # optional port
+    r'(?:/?|[/?]\S+)'
+    r'$', re.IGNORECASE | re.UNICODE)
+
+
 def isurl(text):
-    text = normalize(text)
-    # urlparse does not like port numbers in URLs
-    text = re.sub(r':\d+', '', text)
-    domain = urlparse(text).netloc
-    return validators.domain(domain)
+    return __url_re.match(text) is not None
 
 
 def normalize(url):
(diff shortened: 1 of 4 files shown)


ix5:
Thank you for this PR. In my eyes, we should avoid introducing new dependencies.

The relevant part of the `validators` package is [domain.py](https://github.com/python-validators/validators/blob/506d22bdecd8a049d701a1ef9efcaf9e9f2ba9aa/validators/domain.py).
The package is MIT licensed and we can copy over that function into isso rather than having to load a whole package for a simple regex check.

isso/views/comments.py
```diff
@@ -31,21 +32,12 @@
...
-    r'(?::\d+)?'  # optional port
-    r'(?:/?|[/?]\S+)'
-    r'$', re.IGNORECASE)
-
-
 def isurl(text):
-    return __url_re.match(text) is not None
+    text = normalize(text)
+    # urlparse does not like port numbers in URLs
+    text = re.sub(r':\d+', '', text)
```
jelmer, line 38: This comment seems odd to me - urlparse handles port numbers in URLs fine, so there must be something else going on?

jelmer (changes requested):
Please also add a test for the use case you're trying to address

jelmer:
Do we actually need to check the domain name at all?

ix5:
> Do we actually need to check the domain name at all?

Because IIRC the website is inserted as a link (if given), we should make sure it is valid.

If we skipped the validity check, I'm not sure that the markup escaping would catch e.g. someone entering malicious Javascript `foo</a><span onclick='evil.function()'></span>`.

jelmer:
> > Do we actually need to check the domain name at all?
> 
> Because IIRC the website is inserted as a link (if given), we should make sure it is valid.
> 
> If we skipped the validity check, I'm not sure that the markup escaping would catch e.g. someone entering malicious Javascript `foo</a><span onclick='evil.function()'></span>`.

That seems like an argument for just fixing the markup escaping to me...

schneidr:
> > Do we actually need to check the domain name at all?
> 
> Because IIRC the website is inserted as a link (if given), we should make sure it is valid.
> 
> If we skipped the validity check, I'm not sure that the markup escaping would catch e.g. someone entering malicious Javascript `foo</a><span onclick='evil.function()'></span>`.

So, the goal is to actually check the complete entered URL, not only if the domain is valid, as I assumed?

isso/views/comments.py
schneidr: My bad, the reason for removing the port was not urlparse, it is `validators.domain()` which does not accept `domain:port`. I guess I could clean this up by using `hostname` instead of `netloc`, but if the complete URL is supposed to be validated these lines are most probably not staying anyway.

schneidr: I did add my test case in [test_comments.py](https://github.com/posativ/isso/blob/90019450483c601c1f3dee3de1e973a41679e4d9/isso/tests/test_comments.py#L172).

jelmer:
Sorry for the delay; let's just merge this, since it's clearly an improvement over the current situation. Ideally, we'd be moving away from checking for valid domains to checking for malicious things though.

jelmer (approved):

PR #970, opened 2023-09-06
Add docker-build workflow (#942)

### LINKED ISSUE #942: Please add Multi-CPU architecture for Docker image
Hello, I would kindly request adding support for Multi-CPU architecture on Docker Image with building and pushing images for different architectures during CI/CD workload.

janw:
## Checklist
      

This PR adds an automated docker build using GH actions.

Images are build automatically for the following patterns:

- PR is opened/updated: builds and tags the image but does not push it to the registry (this is done to verify a successful build; [example build](https://github.com/janw/isso/actions/runs/6102764489))
- Push to `master` (i.e. merge a PR): build and tag image with `:latest` ([example build](https://github.com/janw/isso/actions/runs/6102803470))
- Push a new tag (i.e. release new version): build and tag with
  - major version only (i.e. `:0`), and major+minor version (i.e. `:0.14`) ([example build](https://github.com/janw/isso/actions/runs/6102878038/job/16562257652))
  - full version (i.e. `:0.14.1-post.0`) ([example build](https://github.com/janw/isso/actions/runs/6102861527))

Furthermore this includes a multi-arch docker build: the image will be build for linux/amd64, and linux/arm64/v8, arguably the most common platforms for both large production environments and small-scale self-hosted applications.

Example image are available via my fork: https://github.com/janw/isso/pkgs/container/isso

## Why is this necessary?

- Removes necessity to manually build container images, and resolves #942 in the process.

### DIFF
--- a/.github/workflows/docker-build.yaml
+++ b/.github/workflows/docker-build.yaml
@@ -1,0 +1,59 @@
+name: Docker Build
+
+on:
+  push:
+    branches:
+      - "master"
+    tags:
+      - "*"
+  pull_request:
+
+env:
+  build_platforms: ${{ vars.BUILD_PLATFORMS || 'linux/amd64,linux/arm64/v8' }}
+  build_image: ${{ vars.BUILD_IMAGE || 'ghcr.io/isso-comments/isso' }}
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    permissions:
+      packages: write
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+
+      - name: Docker meta
+        id: meta
+        uses: docker/metadata-action@v4
+        with:
+          flavor: |
+            latest=false
+          images: ${{ env.build_image }}
+          tags: |
+            type=ref,event=pr
+            type=semver,pattern={{major}}
+            type=semver,pattern={{major}}.{{minor}}
+            type=semver,pattern={{version}}
+            type=raw,value=latest,enable={{is_default_branch}}
+
+      - name: Login to Github Container Registry
+        if: github.event_name != 'pull_request'
+        uses: docker/login-action@v2
+        with:
+          registry: ghcr.io
+          username: ${{ github.actor }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v2
+
+      - name: Build and push
+        id: docker_build
+        uses: docker/build-push-action@v4
+        with:
+          push: ${{ github.event_name != 'pull_request' }}
+          tags: ${{ steps.meta.outputs.tags }}
+          labels: ${{ steps.meta.outputs.labels }}
+          platforms: ${{ env.build_platforms }}
+
+      - name: Image digest
+        run: echo ${{ steps.docker_build.outputs.digest }}


jelmer:
We probably need to work out where exactly we want the docker images to live, GitHub doesn't make it very easy to publish docker images cross-repo.

@ix5 do we want to move the main repo to the isso-comments org at some point

BBaoVanC:
What other repo is there, why not just publish the images to here?

jelmer:
This PR uses the ghcr.io/isso-comments namespace, and I think the plan is also to move this repo to GitHub.com/isso-comments/isso eventually

ix5:
I talked to @posativ about moving the repo over to the org, but can't quite remember what prevented us from doing that.

PR itself LGTM.

.github/workflows/docker-build.yaml
```diff
@@ -0,0 +1,59 @@
...
+      - name: Docker meta
+        id: meta
+        uses: docker/metadata-action@v4
+        with:
+          flavor: |
+            latest=false
+          images: ${{ env.build_image }}
+          tags: |
+            type=ref,event=pr
+            type=semver,pattern={{major}}
```
ix5, line 33: Where are these values (`{{major}}`, `{{version}}`) pulled from? `git` tags?

janw: That's correct, it's based on semver-formatted tags. More details here: https://github.com/docker/metadata-action#typesemver

ix5:
Just to make it clear, this is blocked until we have a decision about moving the repo over to the org.

jelmer:
> Just to make it clear, this is blocked until we have a decision about moving the repo over to the org.

could we just publish to posativ/isso in the mean time?

ix5:
> > Just to make it clear, this is blocked until we have a decision about moving the repo over to the org.
> 
> could we just publish to posativ/isso in the mean time?

That would be fine by me, but I don't have access to the `Settings` repo tab (for `posativ/isso`) to add a deploy key as a secret.

I _do_ have access to the [isso-comments](https://github.com/isso-comments) org and could add the deploy key as an accepted one, just as I have done for the GitHub pages repo (see [sso-comments/isso-comments.github.io](https://github.com/isso-comments/isso-comments.github.io)).

So either way, we're blocked until Martin finds time to act on this.

janw:
Just checking in, is there anything you need from me to help move this forward? I'd be happy to assist in any way necessary.

jelmer:
Thanks; I think we're basically just waiting from a response from @posativ

jelmer:
The repo has now been migrated, and references are being updated in #996 so let's land this!

.github/workflows/docker-build.yaml
```diff
@@ -0,0 +1,59 @@
...
+            type=semver,pattern={{major}}.{{minor}}
+            type=semver,pattern={{version}}
+            type=raw,value=latest,enable={{is_default_branch}}
```
ix5, line 36: The way I read this, the action rebuilds and pushes the `:latest` image on every push to the default branch (=`master`).

ix5 (commented):
I hadn't realized this at the time, but this change means we're re-building the `:latest` docker tag on _every_ push to `master`.

Up until this change, I had handled the docker image policy as `:latest` for the latest tagged release (`0.13` at the time IIRC) and `:dev` for images from the `master` branch.

Could we please change back to `:latest` being the latest *release*, not "bleeding edge" `master` branch?

jelmer:
I think it's pretty common in docker that latest just refers to whatever the latest thing that was built and pushed was; the way the latest tag works in docker this is pretty common hard to avoid.

I think that if we want to have a docker tag for the latest release, we should just create a new ":release" tag and update that accordingly.

That said, I think it's good practice for people to specify an explicit version (v1.0) as a tag so that their service don't get silently upgraded whenever they restart their services, including any regressions that may cause.

PR #976, opened 2023-11-08
feature: add support for log messages formatting

HorlogeSkynet:
## Checklist
      

Add ability to configure log messages format through configuration.

## Why is this necessary?

`%(asctime)s` can be interesting to add to timestamp log messages.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -14,7 +14,12 @@ New Features
 Breaking Changes
 ^^^^^^^^^^^^^^^^
 
-- TBD
+- Change default log messages formatting and make it configurable (`#976`_, HorlogeSkynet)
+
+    [general]
+    log-format = %(asctime)s:%(levelname)s:%(name)s:%(message)s
+
+.. _#976: https://github.com/posativ/isso/pull/976
 
 Bugfixes & Improvements
 ^^^^^^^^^^^^^^^^^^^^^^^
--- a/docs/docs/reference/server-config.rst
+++ b/docs/docs/reference/server-config.rst
@@ -49,6 +49,7 @@ Here are the **default values** for this section:
     max-age = 15m
     notify = stdout
     log-file =
+    log-format = %(asctime)s:%(levelname)s:%(name)s:%(message)s
     gravatar = false
     gravatar-url = https://www.gravatar.com/avatar/{}?d=identicon&s=55
     latest-enabled = false
@@ -117,6 +118,17 @@ log-file
     Log console messages to file instead of standard out.
 
     Default: (empty)
+
+log-format
+    Format string for console messages logged to file (see ``log-file`` option).
+    Also see Python `LogRecord attributes`_ documentation for the full list of
+    attributes.
+    Please do note this option is interpolated by Python runtime and may lead
+    to security issues if not trusted.
+
+    Default: %(asctime)s:%(levelname)s:%(name)s:%(message)s
+
+.. _LogRecord attributes: https://docs.python.org/3/library/logging.html#logrecord-attributes
 
 gravatar
     When set to ``true`` this will add the property "gravatar_image"
--- a/isso/__init__.py
+++ b/isso/__init__.py
@@ -281,6 +281,9 @@ def main():
     if conf.get("general", "log-file"):
         handler = logging.FileHandler(conf.get("general", "log-file"))
 
+        if conf.get("general", "log-format"):
+            handler.setFormatter(logging.Formatter(conf.get("general", "log-format")))
+
         logger.addHandler(handler)
         logging.getLogger("werkzeug").addHandler(handler)
 
--- a/contrib/isso-dev.cfg
+++ b/contrib/isso-dev.cfg
@@ -16,6 +16,7 @@ max-age = 15m
 notify = stdout
 reply-notifications = false
 log-file =
+log-format = %(asctime)s:%(levelname)s:%(name)s:%(message)s
 latest-enabled = true
 
 [admin]
--- a/isso/isso.cfg
+++ b/isso/isso.cfg
@@ -51,6 +51,10 @@ reply-notifications=false
 # Log console messages to file instead of standard output.
 log-file =
 
+# Format string for console messages logged to file (see Python
+# ``logging.Formatter``).
+log-format = %(asctime)s:%(levelname)s:%(name)s:%(message)s
+
 # adds property "gravatar_image" to json response when true
 # will automatically build md5 hash by email and use "gravatar_url" to build
 # the url to the gravatar image


ix5:
This feature looks interesting, but I have a few thoughts:
- I'm not entirely convinced it is necessary
- At the moment, the linked [Python docs page](https://docs.python.org/3/library/logging.html#logging.Formatter) doesn't really tell you much, maybe you could add a few examples or mention the available format strings (`asctime`, `name`, `message`, `levelname` etc?)
- Security risks; though this would require access to setting the config, format strings are a source of bugs/incidents[1](https://lucumr.pocoo.org/2016/12/29/careful-with-str-format/), [2](https://log4.sh/). I'm not the right person to judge this, perhaps others might chime in how to mitigate those risks?

HorlogeSkynet:
Thanks for reviewing this.

1. It is not strictly "necessary", but timestamped logs are much more interesting (not to say "needed" in some cases). If this patch is rejected, Isso _should_ change its default logs format to include (at least) date and time.

2. That can be done, indeed.

3. To exploit this you need to tweak the configuration. On a regular setup, being able to do so would require same privileges as modifying code. So this risk is affordable, IMHO.

Bye 👋

HorlogeSkynet:
Hey ! Any chance to get this merged before next release ? :slightly_smiling_face:

isso/__init__.py
```diff
@@ -281,6 +281,9 @@ def main():
     if conf.get("general", "log-file"):
         handler = logging.FileHandler(conf.get("general", "log-file"))
 
+        if conf.get("general", "log-format"):
+            handler.setFormatter(logging.Formatter(conf.get("general", "log-format")))
```
ix5, line 285: 1. What happens when the log format is invalid?
2. Why set the log format only for file logs, instead of for stdout as well?

docs/docs/reference/server-config.rst
```diff
@@ -118,6 +119,13 @@ log-file
 
     Default: (empty)
 
+log-format
+    Format string for console messages logged to file (see Python
```
ix5, line 123: 1. Please be more clear that `log-file` needs to be enabled for this to work
2. Please list all available fields and what they do here

CHANGES.rst
```diff
@@ -9,9 +9,14 @@ New Features
 
 - notifications: Implement 'List-Unsubscribe' header (`#906`_, ix5)
 - css: Grey styling for comment code, code blocks (`#907`_, ix5)
+- add support for log messages formatting (`#976`_, HorlogeSkynet)
```
ix5, line 12: Nit: Please capitalize first word

ix5:
Left some review comments inline. Please also add a few tests, our coverage should *increase*, not *decrease* with time.

I'm also partial to changing the default log format to include timestamps. That needs to be communicated, of course.

As for this:
> To exploit this you need to tweak the configuration. On a regular setup, being able to do so would require same privileges as modifying code. So this risk is affordable, IMHO.

It's not necessarily about an attacker already having gained access to the config file, but rather an inexperienced admin changing the defaults (perhaps by including the user-submitted text verbatim in the logs) and thus opening themselves up to vulnerabilities.
At the very least, we should have a small overview of logging format vulnerabilities that might affect Isso here.

ix5 (changes requested):
Please see inline comments

isso/__init__.py
HorlogeSkynet: > 1. What happens when the log format is invalid?

IIRC, there is a traceback when logger tries to emit a message.

> 2. Why set the log format only for file logs, instead of for stdout as well?

Because we are usually interested in persisted logs. And because there was only a  `FileHandler` there (and I didn't want to tweak formatting _globally_).

docs/docs/reference/server-config.rst
HorlogeSkynet: > 1. Please be more clear that `log-file` needs to be enabled for this to work

I've added something.

> 2. Please list all available fields and what they do here

There are way too many, I've opted for a proper link to upstream documentation.

HorlogeSkynet:
> Left some review comments inline. Please also add a few tests, our coverage should _increase_, not _decrease_ with time.

The only code that is modified is the entry point, which isn't actually tested at all.

> I'm also partial to changing the default log format to include timestamps. That needs to be communicated, of course.

The default log format is **not** changed. This patch _allows_ changing it.

> As for this:
> 
> > To exploit this you need to tweak the configuration. On a regular setup, being able to do so would require same privileges as modifying code. So this risk is affordable, IMHO.
> 
> It's not necessarily about an attacker already having gained access to the config file, but rather an inexperienced admin changing the defaults (perhaps by including the user-submitted text verbatim in the logs) and thus opening themselves up to vulnerabilities. At the very least, we should have a small overview of logging format vulnerabilities that might affect Isso here.

I've added another note to the documentation to emphasize that setting this option is not harmless.

Thanks, bye :wave:

jelmer (changes requested):
As commented, I'm skeptical this is a useful feature - especially in light of isso's goal to have reasonable defaults and minimal.

That said, if there is enough support for including this I think there should at least be some tests for it.

HorlogeSkynet:
Hello @jelmer 👋

> As commented, I'm skeptical this is a useful feature - especially in light of isso's goal to have reasonable defaults and minimal.

If I may, as commented, this patch does **not** change current behavior (to address "reasonable defaults") and is actually 2-line long (to address "minimal").

> That said, if there is enough support for including this I think there should at least be some tests for it.

Having logs with proper timestamps is required by law in some countries to answer legal requests (to address "enough support") and I'd love to add a test case to ensure standard library works as expected, if Isso entrypoint were actually tested at all.

Bye 🙏

jelmer:
I think that's an argument for improving the default log format rather than making the log format configurable.

The change isn't minimal, because we now have 
 more code paths - including e.g. handling invalid log format expressions.

HorlogeSkynet:
Hello @jelmer

> I think that's an argument for improving the default log format rather than making the log format configurable.

You will find the branch rebased and updated in order to change the default log format (changelog has been modified accordingly).

> The change isn't minimal, because we now have
> more code paths - including e.g. handling invalid log format expressions.

We don't necessarily have to : if an administrator replaces the (now default) log format by another somewhat broken, Python logging library will emit tracebacks (prefixed by `--- Logging error ---`) whereas Isso will continue to run smoothly.

Bye :wave:

jelmer (changes requested):
Thanks for the update - but that still doesn't address my original comments about making this configurable without adding appropriate tests.

HorlogeSkynet:
> Thanks for the update - but that still doesn't address my original comments about making this configurable without adding appropriate tests.

Hello @jelmer 👋

As I said, many other configuration options (as `log-file`) are not tested. Actually, entry point is not tested at all.
I don't think it is a good thing either, but I'm not sure this issue should be addressed here, as entry point testing is a fully-separate concern.
Also, I fail to see how "testing standard library" downstream would be interesting (as written before, wrong formatting strings will cause explicit exceptions in standard output).

Bye 🙏

jelmer:
The fact that not everything is tested (although @ix5 has made massive strides improving overall test coverage) isn't a good reason for not adding tests for new features (and reducing our overall test coverage).

It would be good to test that invalid log settings don't cause isso to crash, but another thing to test would simply be that the option actually works (i.e. that the log format is actually applied when setting the options). If you look at some of the other PRs open against isso, they all come with tests.

HorlogeSkynet:
> The fact that not everything is tested (although @ix5 has made massive strides improving overall test coverage) isn't a good reason for not adding tests for new features (and reducing our overall test coverage).

I fully agree (see my previous message). If an existing tests suite for Isso entry point would have existed, I would have been glad to extend it according to this "feature".
Re-increasing code coverage (for this conditional branch) could be the subject of another patch, implementing entry point tests.

> It would be good to test that invalid log settings don't cause isso to crash, but another thing to test would simply be that the option actually works (i.e. that the log format is actually applied when setting the options).

I guess such tests got their place in standard library (do they already exist ?). The option actually works, as it runs in production for me since 2023-11-08.

> If you look at some of the other PRs open against isso, they all come with tests.

Indeed, they don't modify/target the entry point.

ix5:
The utility of this change is still nebulous to me. I'd be open to changing isso's default log format (to file, that is, gunicorn already differs anyway) to expose more fields or be better readable. IMO, we haven't made any promises to users regarding the format (stability) of logs so far so a change wouldn't hurt anyone.

I hate to see the collective effort wasted on discussing this PR and really don't want to dissuade the author from contributing to this (or other) projects. He surely won't be too happy with the next paragraph and I can see why they'd feel upset.

However, with the author repeatedly refusing to accept maintainer feedback regarding tests I don't see a path forward for this PR and am going to close this. If jelmer feels differently he may reopen.

HorlogeSkynet:
Hello @ix5 and thanks for taking a decision :pray:

> The utility of this change is still nebulous to me.

This has been addressed above (TL; DR : legal issues regarding metadata ; In the end, is there a software out there writing logs lacking of a proper timestamp ?).

> I hate to see the collective effort wasted on discussing this PR and really don't want to dissuade the author from contributing to this (or other) projects.

:+1: 

I really wonder though how this would "dissuade" me from contributing to OSS. Projects usually approve my patches, not this one, and that's unfortunate (after 6 months, 2 rebases/reworks + 20 messages). But you are the maintainers so you decide.

> I don't see a path forward for this PR

... even accepting that entry point and `log-file` option are already not tested, so `log-format` does not either require to be, or writing proper entry point tests and rebase/extend this patch.

---

If some users/administrators happen to get to the bottom of this endless discussion, I'll be maintaining the patch [here](https://gist.github.com/HorlogeSkynet/7052787db2d3e2027a848ed399d824c4), while Isso does not support this feature (as I am legally required to do so).

---

Thanks, bye :wave:

PR #981, opened 2024-01-15
Add error message for improperly configured server.listen

sjrct:
## Checklist
- [x] All new and existindocs/docs/technical-docs/testing.rstg **tests are passing**
      

Add error message for improperly configured `server.listen`

## Why is this necessary?
Otherwise the error message is quite cryptic, what I saw was ultimately:
```
AttributeError: 'int' object has no attribute 'strip'
```

### DIFF
--- a/isso/__init__.py
+++ b/isso/__init__.py
@@ -295,7 +295,7 @@ def main():
         except ImportError:
             run_simple(host, port, make_app(conf), threaded=True,
                        use_reloader=conf.getboolean('server', 'reload'))
-    else:
+    elif conf.get("server", "listen").startswith("unix://"):
         sock = conf.get("server", "listen").partition("unix://")[2]
         try:
             os.unlink(sock)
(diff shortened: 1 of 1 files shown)


isso/__init__.py
```diff
@@ -295,11 +295,14 @@ def main():
...
+    elif conf.get("server", "listen").startswith("unix://"):
         sock = conf.get("server", "listen").partition("unix://")[2]
         try:
             os.unlink(sock)
         except OSError as ex:
             if ex.errno != errno.ENOENT:
                 raise
         wsgi.SocketHTTPServer(sock, make_app(conf)).serve_forever()
+    else:
+        logger.error("server.listen must specif a protocol of http:// or unix://")
```
BBaoVanC, line 307: nit: should say "specify" instead of "specif"

BBaoVanC (approved):
Looks good, I think there's a small typo though

isso/__init__.py
sjrct: Fixed

ix5:
@BBaoVanC if you're satisfied, I would like to merge this.

Issue #979, opened 2023-11-30
NULL text from deleted comment results in error 500


ttencate:
## Checklist

- [x] I am running the latest version. (Using 0.13.0 from pip.)
- [ ] Installing Isso from GitHub from the `master` branch does not fix my issue. (Haven't checked because I don't want to install it on a production website.)
- [x] I have checked the [troubleshooting guide](https://isso-comments.de/docs/guides/troubleshooting/)
- [x] I have searched the [open issues](https://github.com/posativ/isso/issues), but my issue has not already been reported

## What is not working?

I imported comments from Disqus, among which the following:

```xml
  <post dsq:id="2687640654">                                                          
    <id/>
    <message/>
    <createdAt>2016-05-21T08:39:28Z</createdAt>
    <isDeleted>true</isDeleted>
    <isSpam>false</isSpam>
    <author>
      <name/>
      <isAnonymous>true</isAnonymous>
    </author>
    <thread dsq:id="4843794432"/>
  </post>
```

The comment was deleted in Disqus, but has a child that's not deleted, so both have been exported and both were imported.

However, the deleted parent got imported into the Isso `comments` table with `text` set to `NULL`, which causes a crash when retrieved:

```
2023-11-30 15:22:04,646 ERROR: GET /
Traceback (most recent call last):
  File "isso/__init__.py", line 152, in dispatch
    response = handler(request.environ, request, **values)
  File "isso/views/__init__.py", line 43, in dec
    return func(cls, env, req, *args, **kwargs)
  File "isso/views/comments.py", line 894, in fetch
    'replies': self._process_fetched_list(root_list, plain),
  File "isso/views/comments.py", line 953, in _process_fetched_list
    item['text'] = self.isso.render(item['text'])
  File "isso/__init__.py", line 129, in render
    return self.markup.render(text)
  File "isso/utils/html.py", line 104, in render
    return self._render(text)
  File "isso/utils/html.py", line 101, in <lambda>
    self._render = lambda text: sanitizer.sanitize(parser(text))
  File "isso/utils/html.py", line 58, in inner
    rv = md(text).rstrip("\n")
  File "misaka/api.py", line 176, in __call__
    lib.hoedown_buffer_puts(ib, text.encode('utf-8'))
AttributeError: 'NoneType' object has no attribute 'encode'
```

As a workaround, setting `text` to `''` (the empty string) manually through the SQLite command line fixed it for me, and now the comment correctly shows up as "Comment deleted." in the frontend.

[...]

PR #994, opened 2024-02-27
isso: migrate: Handle deleted comments in Disqus migration

pkvach:
## Checklist
- [x] I have added an entry to CHANGES.rst because this is a user-facing change or an important bugfix
      

- `isso/migrate.py`: Improved handling of deleted comments during Disqus migration by ensuring that deleted comments are imported with an empty message text, enhancing the robustness of the migration process.
- `isso/utils/html.py`: Add conditional check for `None` before Markdown conversion in HTML utility, improving the system's resilience to null input values.
- Updated unit tests, ensuring that the changes are correctly implemented.

## Why is this necessary?

Fixes https://github.com/posativ/isso/issues/979


jelmer:
please complete the checklist in the PR template

pkvach:
Thank you for the review.
I apologize if I got it wrong, as I'm doing my first PRs here. 
Added an entry to CHANGES.rst

jelmer:
I don't have much background on disqus, but what's the point of still importing these comments but then making isso ignore them? Couldn't we just filter out the deleted comments during import or is there some point in keeping them?

pkvach:
During migration, deleted comments are ignored if they have no child comments.
But in case like https://github.com/isso-comments/isso/issues/979, the deleted comment has children. And then we have to migrate it.
This follows the same logic as in the main system: a comment is not completely deleted if it has children.

jelmer:
Is setting the comment text to null the right thing to do in that case though, or we just set it to an empty string at import time (and change the schema to disallow null) ?

pkvach:
In the main scenario, if a comment needs to be marked as deleted and not removed from the database, its text is set to an empty string. 
https://github.com/isso-comments/isso/blob/fc0641fb6b91bf599aa7ae6866e973aa679e06c4/isso/db/comments.py#L285

But in the Disqus migration `null` was set and it produced an error when processing comments for output:
```
2023-11-30 15:22:04,646 ERROR: GET /
  File "isso/__init__.py", line 152, in dispatch
  File "isso/views/__init__.py", line 43, in dec
  File "isso/views/comments.py", line 894, in fetch
    'replies': self._process_fetched_list(root_list, plain),
  File "isso/views/comments.py", line 953, in _process_fetched_list
    item['text'] = self.isso.render(item['text'])
  File "isso/__init__.py", line 129, in render
    return self.markup.render(text)
  File "isso/utils/html.py", line 104, in render
    return self._render(text)
  File "isso/utils/html.py", line 101, in <lambda>
    self._render = lambda text: sanitizer.sanitize(parser(text))
  File "isso/utils/html.py", line 58, in inner
    rv = md(text).rstrip("\n")
  File "misaka/api.py", line 176, in __call__
    lib.hoedown_buffer_puts(ib, text.encode('utf-8'))
AttributeError: 'NoneType' object has no attribute 'encode'
```

jelmer:
Right, what I'm saying is - why not fix this in the disqus migration by having it set message to an empty string when it is null?

pkvach:
Sorry if I misunderstood you, but that's what I made the changes to `migrate.py` for. If the comment text is not specified, set it to an empty string.
Changes in `html.py` to prevent crashing if comment text is null. This is like an extra check in case such comments are already written to the database or will be written in other cases.

jelmer:
> Sorry if I misunderstood you, but that's what I made the changes to `migrate.py` for. If the comment text is not specified, set it to an empty string. Changes in `html.py` to prevent crashing if comment text is null. This is like an extra check in case such comments are already written to the database or will be written in other cases.

Yeah, it's the change in html.py that I'm objecting to. We should rather just initialize the database to mandate that text can't be null rather than adjusting the application to cope with it.

pkvach:
> Yeah, it's the change in html.py that I'm objecting to. We should rather just initialize the database to mandate that text can't be null rather than adjusting the application to cope with it.

That's a good point, thank you.
I reverted the changes on `html.py` and added `NOT NULL` constraint for the `text` field when creating the `comment` table. 
Do you think we need to add some sort of migration for existing databases?

jelmer:
> > Yeah, it's the change in html.py that I'm objecting to. We should rather just initialize the database to mandate that text can't be null rather than adjusting the application to cope with it.
> 
> That's a good point, thank you. I reverted the changes on `html.py` and added `NOT NULL` constraint for the `text` field when creating the `comment` table. Do you think we need to add some sort of migration for existing databases?

That'd be a nice addition (in isso/db.py) but I don't think it's strictly necessary

jelmer (approved):

ix5:
Apologies for the late comment. It appears to me that since this PR changes the DB schema in https://github.com/isso-comments/isso/commit/6dffaa74f1184280f45a8bf64730f3d153c74b05#diff-115e97a46b3ca0f35e7455d6e624550eff6fe684c3354f42545d06b874eba433R41 we ought to update the internal database version and add a migration clause:

https://github.com/isso-comments/isso/blob/6dffaa74f1184280f45a8bf64730f3d153c74b05/isso/db/__init__.py#L66-L123

pkvach:
@ix5 
Thank you for the review.
Should I make a new PR to add changes on migration?

ix5:
> @ix5 Thank you for the review. Should I make a new PR to add changes on migration?

Yes please!

pkvach:
> > @ix5 Thank you for the review. Should I make a new PR to add changes on migration?
> 
> Yes please!

Added a PR with migration: https://github.com/isso-comments/isso/pull/1019

PR #996, opened 2024-03-05
change GitHub organization

posativ:
Change `posativ/isso` to `isso-comments/isso`

Not sure if docs/news.rst or CHANGES.rst should also be edited.


jelmer:
> Not sure if docs/news.rst or CHANGES.rst should also be edited.

I think I can see an argument either way. It would be good for consistency, but also CHANGES a historic record, so might make sense to keep the original lines. It doesn't seem likely that the redirect will break.

So let's keep this as is for now? Other changes LGTM.

jelmer (approved):

BBaoVanC:
> think I can see an argument either way. It would be good for consistency, but also CHANGES a historic record, so might make sense to keep the original lines.

In my opinion the concern should be whether the document is a working historical record, not that text written years ago stays unmodified. I personally would update all the links to the new URL, but given that GitHub would make redirects it doesn't really matter unless something changes with GitHub or something in the future.

jelmer:
> > think I can see an argument either way. It would be good for consistency, but also CHANGES a historic record, so might make sense to keep the original lines.
> 
> In my opinion the concern should be whether the document is a working historical record, not that text written years ago stays unmodified. I personally would update all the links to the new URL, but given that GitHub would make redirects it doesn't really matter unless something changes with GitHub or something in the future.

Your quote seems strip out the last line from my reply, which said pretty much the same thing.

jelmer:
This looks good to me as-is, and hasn't had any more comments so I think we should just land it.

PR #1000, opened 2024-03-17
admin: Add support for comment search by URL

pkvach:
## Checklist
      

Added support for searching comments by URL in the admin interface. This includes changes to the backend to handle comment URL queries and extract comment IDs, and frontend changes to add a search input to the admin page.

- Updated admin CSS to include styles for the new search input and button.

- Modified the fetchall method in comments.py to accept an optional comment_id parameter for fetching a specific comment by ID.

- Updated the admin HTML template to include a search form that allows inputting a comment URL for search.

- Updated the admin method in views/comments.py to handle the comment_url query parameter, parse the URL to extract the comment ID, and fetch the corresponding comment.

## Why is this necessary?

Closes https://github.com/isso-comments/isso/issues/642

### Demo:
image

image

### DIFF
--- a/isso/views/comments.py
+++ b/isso/views/comments.py
@@ -12,7 +12,7 @@ from datetime import datetime, timedelta
 from html import escape
 from io import BytesIO as StringIO
 from os import path as os_path
-from urllib.parse import unquote, urlparse
+from urllib.parse import unquote, urlparse, urlsplit
 from xml.etree import ElementTree as ET
 
 from itsdangerous import SignatureExpired, BadSignature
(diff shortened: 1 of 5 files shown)


isso/views/comments.py
```diff
@@ -1396,10 +1398,25 @@ def admin(self, env, req):
...
+        # Search for a specific comment by URL
+        if comment_url:
+            try:
+                # Parse the comment URL to extract the comment ID from the fragment
+                parsed_url = urlparse(comment_url)
+                fragment = parsed_url.fragment
+                comment_id = int(fragment.split('-')[-1])
+
+                comments = self.comments.fetchall(comment_id=comment_id, limit=1)
+            except Exception:
```
jelmer, line 1412: Please catch more specific exceptions and more specific to what's raising them; "Exception" is too broad and can catch exceptions we don't intend to.

jelmer (approved):

ix5:
If I see this correctly, the search only succeeds if you supply both the **thread** URL as well as a per-thread **comment** id.

What about admins who would want to look up only `http://localhost:8080/demo/`, e.g. the **thread** URL, to moderate all comments for a particular blogpost? Since this was merged already, could you add that in a follow-up PR?

ix5:
> Yes, this PR implements the requested search for specific comments. I agree that we could make it even better by allowing comments to be searchable by thread URL.

Could you create a follow-up PR? I would argue that admins most often would search for threads, not comments.

pkvach:
Added a new PR to search for comments by thread URL: https://github.com/isso-comments/isso/pull/1020

PR #1001, opened 2024-03-21
css: Add CSS variables for better organization and flexibility

pkvach:
## Checklist
      

Added CSS variables to improve theme customization and organization in `isso.css`.

## Why is this necessary?

Closes https://github.com/isso-comments/isso/issues/655

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -9,9 +9,11 @@ New Features
 
 - Add Catalan localisation (`#966`_, welpo)
 - Make <code class="language-$lang"> for syntax highlighting (`#998`_, pkvach)
+- Add CSS variables for better organization and flexibility (`#1001`_, pkvach)
 
 .. _#966: https://github.com/posativ/isso/pull/966
 .. _#998: https://github.com/isso-comments/isso/pull/998
+.. _#1001: https://github.com/isso-comments/isso/pull/1001
 
 Breaking Changes
 ^^^^^^^^^^^^^^^^
--- a/isso/css/isso.css
+++ b/isso/css/isso.css
@@ -1,3 +1,41 @@
+:root {
+    --isso-primary-text-color: #555;
+    --isso-secondary-text-color: #808080;
+
+    --isso-border: 1px solid rgba(0, 0, 0, 0.2);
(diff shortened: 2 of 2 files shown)


jelmer (approved):

ix5:
Nice work, very cleanly done. LGTM as well.

---

One thing though: I'll reiterate what I wrote in 

> I do not have a particularly strong opinion on this either way. But I agree with @ vincentbernat in testing before blindly rewriting everything.
>
> As a data point: https://caniuse.com/css-variables
>
> Browsers from 2017 (Edge) might not support it. IE 10 doesn't support it. Many big-C Corporate/Enterprise people are still on LTS versions, one way or another. Also, Android 4.4 devices are still working perfectly fine.
> Also, switching the CSS would break existing setups with custom CSS, so it'd require a version bump and a notice to users.

This is potentially a breaking change as it drops support for IE10 (not sure whether we were compatible before, but now we definitely aren't). We should communicate that somewhere (changelog) and also drop the claim from the homepage ([docs/index.html](https://github.com/isso-comments/isso/blob/c257a4fe9490632bc29abc1dc294847bab06fcc7/docs/index.html#L35)) about IE10 compatibility.

pkvach:
Thank you for your attention to this matter.

I understand that there may be differing opinions regarding whether to continue supporting a browser that is still in use but no longer receives updates from its developer, leaving it vulnerable to security risks.

If needed, I can help with restoring support for IE 10 after this PR change.

ix5:
> Thank you for your attention to this matter.
> 
> I understand that there may be differing opinions regarding whether to continue supporting a browser that is still in use but no longer receives updates from its developer, leaving it vulnerable to security risks.
> 
> If needed, I can help with restoring support for IE 10 after this PR change.

I might have miscommunicated here, IE10 support is not that important to me, but rather our claims to be compatible when we aren't anymore.

pkvach:
Great, thanks for clarifying.
Should I correct the documentation where IE10 support is listed? 
And maybe add a note to the “Changes.rst” file.

ix5:
> Great, thanks for clarifying. Should I correct the documentation where IE10 support is listed? And maybe add a note to the `CHANGES.rst` file.

Already done https://github.com/isso-comments/isso/pull/1022

PR #1007, opened 2024-04-16
utils/html: Provide full control of allowed HTML elements via the configuration file

### LINKED ISSUE #751: Isso config [markup] can't work
I found an issue about the markup config, any suggestions would be appreciate.

Step1, add below code to isso server config,
```[markup]
options = strikethrough, superscript, autolink
allowed-elements = img, blockquote, br, code, del, em, h1, h2, h3, h4, h5, h6, hr, ins, ul, li, ol, p, pre, strong, table, tbody, td, th, thead
allowed-attributes = src
```

Step2, In the website where you refer isso as comment fn, add below comment:
`<a href=http://evil.com>clickmetochangethispassword</a>`

Result: the comment was successfully rendered.

Expect: transform the a markup.

pkvach:
## Checklist
      

Add `allowed-html-elements` configuration option
- Defines the exact set of HTML tags permitted in rendered output
- Replaces the built-in default list (unlike `allowed-elements`, which appends to it)

Deprecate `allowed-elements`
- When both options are set, its tags are still merged on top of
  `allowed-html-elements` for backwards compatibility
- A deprecation warning is logged when it is used

Update bundled configuration files
- Switch to the new `allowed-html-elements` option
- Add `tr` to the default allowed element list

## Why is this necessary?

Fixes https://github.com/isso-comments/isso/issues/751

### DIFF
--- a/contrib/isso-dev.cfg
+++ b/contrib/isso-dev.cfg
@@ -38,7 +38,7 @@ reply-to-self = true
 [markup]
 options = autolink, fenced-code, no-intra-emphasis, strikethrough, superscript
 flags =
-allowed-elements =
+allowed-html-elements =
 allowed-attributes =
 
 [hash]
--- a/docs/docs/reference/server-config.rst
+++ b/docs/docs/reference/server-config.rst
@@ -438,7 +438,7 @@ For a more detailed explanation, see :doc:`/docs/reference/markdown-config`.
     renderer = mistune
     options = strikethrough, superscript, autolink, fenced-code
     flags =
-    allowed-elements =
+    allowed-html-elements =
     allowed-attributes =
 
 renderer
@@ -470,7 +470,7 @@ allowed-elements
 
     By default, only ``a``, ``blockquote``, ``br``, ``code``, ``del``, ``em``,
     ``h1``, ``h2``, ``h3``, ``h4``, ``h5``, ``h6``, ``hr``, ``ins``, ``li``,
-    ``ol``, ``p``, ``pre``, ``strong``, ``table``, ``tbody``, ``td``, ``th``,
+    ``ol``, ``p``, ``pre``, ``strong``, ``table``, ``tbody``, ``tr``, ``td``, ``th``,
     ``thead`` and ``ul`` are allowed.
 
     For a more detailed explanation, see :doc:`/docs/reference/markdown-config`.
@@ -482,10 +482,28 @@ allowed-elements
        mean that ``br, code, del, ...`` and all other default allowed tags are
        still allowed. You can only add *additional* elements here.
 
-       It is planned to change this behavior, see
-       `this issue <https://github.com/isso-comments/isso/issues/751>`_.
+       To specify a list of *only* allowed elements, use the
+       ``allowed-html-elements`` option instead.
 
     Default: (empty)
+
+    .. deprecated:: 0.14.1
+
+       Superseded by ``allowed-html-elements``, which gives full control over
+       the list of allowed elements instead of only appending to it. If both
+       options are set, elements listed here are still added on top of
+       ``allowed-html-elements`` for backwards compatibility, but a warning
+       is logged. Please migrate to ``allowed-html-elements``.
+
+allowed-html-elements
+
+    **Only** allow the specified HTML tags in the generated output, comma-separated.
+    If ``allowed-elements`` is also set, its elements are added on top of this
+    list (with a deprecation warning logged); migrate them here instead.
+
+    Default: (empty)
+
+    .. versionadded:: 0.14.1
 
 allowed-attributes
     **Additional** HTML attributes (independent from elements) to allow in the
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -12,12 +12,14 @@ New Features
 - Add no-cache headers to fix stale comment display (`#1110`_, pkvach)
 - Add webhook support for comment events (`#1123`_, p1slave)
 - Update comment counts in the admin interface dynamically after moderation. (`#1113`_, pkvach)
+- Provide full control of allowed HTML elements via the configuration file (`#1007`_, pkvach)
 
 .. _#1111: https://github.com/isso-comments/isso/pull/1111
 .. _#1117: https://github.com/isso-comments/isso/pull/1117
 .. _#1110: https://github.com/isso-comments/isso/pull/1110
 .. _#1123: https://github.com/isso-comments/isso/pull/1123
 .. _#1113: https://github.com/isso-comments/isso/pull/1113
+.. _#1007: https://github.com/isso-comments/isso/pull/1007
 
 0.14.0 (2026-03-26)
 --------------------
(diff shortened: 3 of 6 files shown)


contrib/isso-dev.cfg
```diff
@@ -37,8 +37,9 @@ reply-to-self = true
 [markup]
 options = autolink, fenced-code, no-intra-emphasis, strikethrough, superscript
 flags =
-allowed-elements =
-allowed-attributes =
+allowed-elements = a, p, hr, br, ol, ul, li, pre, code, blockquote, del, ins,
```
ix5, line 40: Please add `tr` to the allowed elements in a separate commit, it's not reasonable that we allow all other table elements but this one

docs/docs/reference/server-config.rst
```diff
@@ -428,37 +428,19 @@ flags
     .. versionadded:: 0.12.4
 
 allowed-elements
-    **Additional** HTML tags to allow in the generated output, comma-separated.
-
-    By default, only ``a``, ``blockquote``, ``br``, ``code``, ``del``, ``em``,
-    ``h1``, ``h2``, ``h3``, ``h4``, ``h5``, ``h6``, ``hr``, ``ins``, ``li``,
-    ``ol``, ``p``, ``pre``, ``strong``, ``table``, ``tbody``, ``td``, ``th``,
-    ``thead`` and ``ul`` are allowed.
+    HTML tags to allow in the generated output, comma-separated.
```
ix5, line 431: Use `versionchanged` here

isso/utils/html.py
```diff
@@ -17,25 +17,17 @@ def allow_attribute_class(tag, name, value):
...
 
         # allowed attributes for tags
         self.attributes = {
-            "table": ["align"],
-            "a": ["href"],
-            "code": Sanitizer.allow_attribute_class,
             "*": attributes
         }
 
+        # If "code" elements are allowed, allow "language-*" CSS classes for syntax highlighting
```
ix5, line 27: Could you add a few more quality-of-life improvements to `Markup` in a separate commit below?

E.g. `if highlight` -> `allow <mark>`

ix5 (changes requested):
Thank you for tackling this. I will write my more detailed thoughts on this down later.

For now:
1. Keep in mind that this is a massive footgun for people who used this feature as intended and now all their styling is gone.
2. The whole markdown rendering configuration is extremely un-intuitive and error prone. We essentially only pass through tunables to `misaka` and `bleach` instead of providing the admin with  sane options.
3. People expecting GitHub-flavored Markdown (GHFM) to just work are massively let down.

isso/utils/html.py
pkvach: I added checks for the `highlight` and `underline` extensions.

pkvach:
> Thank you for tackling this. I will write my more detailed thoughts on this down later.
> 
> For now:
> 
> 1. Keep in mind that this is a massive footgun for people who used this feature as intended and now all their styling is gone.
> 2. The whole markdown rendering configuration is extremely un-intuitive and error prone. We essentially only pass through tunables to `misaka` and `bleach` instead of providing the admin with  sane options.
> 3. People expecting GitHub-flavored Markdown (GHFM) to just work are massively let down.

Thanks for the thoughtful feedback! I appreciate you taking the time to consider the impact on existing users.

Backwards compatibility is a valid concern. If you have any ideas on how we might mitigate disruption or implement the changes more seamlessly, I'd be happy to hear them.

ix5:
> Backwards compatibility is a valid concern. If you have any ideas on how we might mitigate disruption or implement the changes more seamlessly, I'd be happy to hear them.

My biggest concern with this particular PR would be users who have copied a sample config file a few years ago that contained `allowed-elements =` and now nothing renders at all.

I have a few ideas, none of them perfect:

1. Use a different parameter name (e.g. `allowed-html-elements`) for the "new" behavior and use it only if the new config key is non-empty
2. Compare user-supplied `allowed-elements` against the default list and warn (in the the server logs) if the user-supplied list is missing default items, or if it is significantly shorter (e.g. it only contains 3 items, then the user most likely wanted the old behavior)

The config parser silently merging the default config with a user-supplied one also doesn't make things easier here.

---

As for general thoughts on our markdown rendering: Do we really want to make incremental improvements to this pile of hacks or would energy better be spent overhauling the thing and making it radically more user-friendly?

I'm happy to review and merge PRs for incremental fixes, but it feels more like treading water than tackling the real issue. Note also that both `misaka` as well as `bleach` ([see here](https://github.com/mozilla/bleach/issues/698), thanks to https://github.com/isso-comments/isso/issues/936#issuecomment-2083351957) are deprecated and might not survive the jump to Python 3.13 or whatever comes next.

pkvach:
Thank you for your thoughts on this.

> 1. Use a different parameter name (e.g. allowed-html-elements) for the "new" behavior and use it only if the new config key is non-empty

I've been thinking about that option as well, and I'm leaning more towards it. I will try to move in that direction.

pkvach:
Made changes to PR with alternative option: added new configuration option `strictly-allowed-html-elements` to specify only allowed HTML tags in generated output.

ix5:
Hey @pkvach, your solution is well-written code-wise, but I still would like to receive some (more) input from actual users regarding their experiences and expectations with markdown/HTML allowlist handling.

How do other commenting systems (Commento, Remark42, ...) handle element/attribute whitelisting, if at all?

pkvach:
Thanks for the feedback, @ix5 .
I agree that we can take our time with these changes, as there is no rush here.
As for other commenting systems, I can do some research to see how they handle whitelisting.

pkvach:
It seems that both “Remark42” and “Commento” systems do not provide the ability to specify a list of allowed elements in the configuration. 

Both use a processor for Markdown:
https://github.com/russross/blackfriday

https://github.com/adtac/commento/blob/a4c5cc8b9e5749dd50ef01f94c08dd7944a6d9c1/api/markdown_html.go#L8

https://github.com/umputun/remark42/blob/f7ba43e5f159d7431c2f3a7000f21d374be541b3/backend/app/store/formatter.go#L128

Remark42 also uses a sanitizer:
https://github.com/microcosm-cc/bluemonday

https://github.com/umputun/remark42/blob/f7ba43e5f159d7431c2f3a7000f21d374be541b3/backend/app/store/comment.go#L119

ix5:
Thank you for your extensive research! At the moment, I'm too much occupied with IRL things to delve deeper into this and make a decision. Maybe one of @jelmer, @BBaoVanC, @stefangehn @fluffy-critter would care to comment and try to form some kind of consensus.

stefangehn (commented):
I like the change in general but keeping both configuration variants around looks confusing to me. I would rather suggest to deprecate the old setting and replace it with the new one.

what about this logic:
- introduce `allowed-html-elements` key and have a default value for it based on the hardcoded list in the middle of the code
- add the list of `allowed-elements` *on top* of what the 
`allowed-html-elements` already provided
- warn when any items from `allowed-elements` were added, pointing users to migrate their config to `allowed-html-elements` in the future
- drop the `allowed-elements` key from the default/example configuration file so no new users start using it

jelmer:
Hey @pkvach , are you still pursuing this?

pkvach:
Hey, thanks for following up! 
I apologize for missing the latest ideas from stefangehn on this. I'll take a closer look at them and work on implementing the improvements.

PR #1011, opened 2024-04-18
Drop reference to matrix bridge

### LINKED ISSUE #989: matrix bridge is gone
Joining #matrix:libera.chat no longer works, presumably because libera and matrix have deportalled.

https://matrix.org/blog/2023/07/deportalling-libera-chat/

jelmer:
See https://matrix.org/blog/2023/07/deportalling-libera-chat/

Fixes #989

### DIFF
--- a/README.md
+++ b/README.md
@@ -64,8 +64,7 @@ Refer to the docs for
 
 ### Help
 
-- Join `#isso` via [Matrix](https://matrix.to/#/#isso:libera.chat) or via IRC on
-  [Libera.Chat](https://libera.chat/)
+- Join `#isso` via IRC on [Libera.Chat](https://libera.chat/)
 - Ask a question on [GitHub Discussions](https://github.com/isso-comments/isso/discussions).
 
 ## License


ix5 (approved):

ix5:
> The matrix bridge wasn't any more active though. I think either way, we should pick one chat mechanism rather than advertising more (and bifurcating the community).

Not disagreeing with you about the bridge, that one's been too buggy and discontinued as you already mentioned.

So far, the most helpful users-helping-users avenue seems to be the GH Discussions area, and I suspect this is due to a) searchability/Google influx, b) persistence (without a bouncer) and c) ease-of-use. Other OSS projects seem to have success with Gitter, Discord or Slack, with more libre-tending projects favoring IRC/Matrix.

I feel it's a bit disingenuous to point people to the IRC when they have no realistic "guarantee"/expectation of a) whether and b) when someone will even answer them. At least with services that provide a room history, a user can gauge the activity of a room/channel before shouting into the void.

Sorry for not providing a clear solution, just my thoughts on the matter.

jelmer:
I feel like that's a general issue with *any* realtime support channels, nothing specific to IRC/Matrix. GitHub discussions aren't realtime. Sometimes there are people around on IRC that can help, sometimes there are not. Advertising a single channel for realtime discussions at least means the likelihood of people being able to find each other is higher.

PR #1014, opened 2024-04-21
Docker, docs: Document and implement changed image tagging policy

ix5:
## Checklist
      

- `:latest` is now used for auto-pushed images off of `master`, while `:release` will follow the latest tagged stable release.
- Use `org.opencontainers.*` docker `LABEL` metadata to connect images to our GitHub source

## Why is this necessary?

https://github.com/isso-comments/isso/issues/1012

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -22,7 +22,12 @@ New Features
 Breaking Changes
 ^^^^^^^^^^^^^^^^
 
-- TBD
+- Rework the Docker image tagging scheme: ``isso:latest`` is now rebuilt on
+  every push to ``master``, while ``isso:release`` points to the latest stable
+  release (`#970`_, janw). Previously, ``:latest`` pointed to the latest stable
+  (tagged) version
+
+.. _#970: https://github.com/isso-comments/isso/pull/970
 
 Bugfixes & Improvements
 ^^^^^^^^^^^^^^^^^^^^^^^
--- a/Dockerfile
+++ b/Dockerfile
@@ -71,6 +71,10 @@ FROM python:3.10-alpine AS isso
 WORKDIR /isso/
 COPY --from=isso-builder /isso/ .
 
+LABEL org.opencontainers.image.source=https://github.com/isso-comments/isso
+LABEL org.opencontainers.image.description="Isso – a commenting server similar to Disqus"
+LABEL org.opencontainers.image.licenses=MIT
+
 # Clean up
 RUN rm -rf /var/apk/cache/* /tmp/* /var/tmp/*
 
--- a/README.md
+++ b/README.md
@@ -45,9 +45,19 @@ the [the full documentation](https://isso-comments.de/docs/).
 
 ## Docker
 
-A Docker image with the latest stable release is provided at
-`ghcr.io/isso-comments/isso:latest`. See
+> [!NOTE]  
+> The Docker image tagging scheme for stable releases was changed from `:latest`
+> to `:release` as of March 2024
+> ([#970](https://github.com/isso-comments/isso/pull/970), [#1012](https://github.com/isso-comments/isso/issues/1012))
+
+A [Docker image](https://github.com/isso-comments/isso/pkgs/container/isso) with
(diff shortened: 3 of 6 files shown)


CHANGES.rst
```diff
@@ -22,7 +22,12 @@ New Features
 Breaking Changes
 ^^^^^^^^^^^^^^^^
 
-- TBD
+- Rework the Docker image tagging scheme: ``isso:latest`` is now rebuilt on
+  every push to ``master``, while ``isso:release`` points to the latest stable
+  release (`#970`_, janw). Previously, ``:latest`` pointed to the latest stable
```
jelmer, line 27: Was this ever actually true?

Dockerfile
```diff
@@ -71,6 +71,10 @@ FROM python:3.10-alpine AS isso
 WORKDIR /isso/
 COPY --from=isso-builder /isso/ .
 
+LABEL org.opencontainers.image.source=https://github.com/isso-comments/isso
```
jelmer, line 74: This isn't mentioned in the pr description

docs/docs/reference/installation.rst
```diff
@@ -149,12 +149,30 @@ two options for running a Docker container:
 a) Official Docker image
 ^^^^^^^^^^^^^^^^^^^^^^^^
 
+.. attention::
+
+   The Docker image tagging scheme for stable relases was changed from ``:latest``
```
jelmer, line 154: releases

README.md
```diff
@@ -45,10 +45,20 @@ the [the full documentation](https://isso-comments.de/docs/).
 
 ## Docker
 
-A Docker image with the latest stable release is provided at
-`ghcr.io/isso-comments/isso:latest`. See
+> [!NOTE]  
+> The Docker image tagging scheme for stable relases was changed from `:latest`
```
jelmer, line 49: I don't think this was ever actually true, so it doesn't seem useful to document that as previous behaviour.

CHANGES.rst
ix5: Yup, that is back from when I manually released docker images to the `ghcr.io/isso-comments` namespace. Not particularly knowing any better, using `:lastest` was the tagging strategy I implemented for releases. I used `:dev` for images off of `master`.

Dockerfile
ix5: Amended the PR description

docs/docs/reference/installation.rst
ix5: Also fixed in README, thanks

README.md
ix5: See above, that was the tagging policy I implemented and which was previously documented as such.

> A Docker image with the latest stable release is provided at `ghcr.io/isso-comments/isso:latest` [...]

It seems at least one user already relied on this, so I deem it prudent to document this change rather than silently `s/:latest/:release` in the docs.

CHANGES.rst
jelmer: Ah, fair enough

jelmer (changes requested):
LGTM, one comment:

Please update the :release tag from the CI - I don't think the current PR does that.

ix5:
> LGTM, one comment:
> 
> Please update the :release tag from the CI - I don't think the current PR does that.

I've added https://github.com/isso-comments/isso/pull/1014/commits/1544b34691ecfc2b511123951641a6089014b529 `.github/docker-build: Tag ':release' on git tags` which additionally tags the newly built docker image with `:release` if the GitHub Actions event type is `tag` (`github.ref_type == 'tag'`).

@jelmer, you will have to remember this in case you want to push `debian/*`-namespaced tags. I'm not in the mood to dig further into GitHub Actions' YAML-based expression evaluation to make this action smarter (e.g. only tag the docker image if tag matches semver AND the version is the "highest" out of all tags).

ix5:
CI is failing for the compare-screenshots step because I rebuilt and pushed the `testbed` image for node-related stuff - will fix in another PR.

jelmer (approved):

PR #1015, opened 2024-04-23
Python 3.12 compatibility; docker, venv/virtualenv

ix5:
## Checklist
      

Compatibility for Python 3.12 and a few stray nits

* Mentions needed `virtualenv` package in docs, make pip and virtualenv related commands more consistent
* Require `setuptools` in `setup.py` (mostly because of `pkg_resources`)
* docker: Use buildkit in another instance of the `Makefile`, remove version from `compose` file, improve `Dockerfile` formatting
* Mark Python 3.12 support in `setup.py` and use it in `Dockerfile`, GitHub Actions

## Why is this necessary?

Python 3.12 changes: [What's new](https://docs.python.org/3/whatsnew/3.12.html)

> [gh-95299](https://github.com/python/cpython/issues/95299): Do not pre-install `setuptools` in virtual environments created with `venv`. This means that `distutils`, `setuptools`, `pkg_resources`, and `easy_install` will no longer available by default; to access these run `pip install setuptools` in the activated virtual environment.

- We were using `python -m venv` and `virtualenv` interchangeably in the docs
- In Python 3.12, `python -m venv` seemingly no longer pulls in `pip` and `setuptools`, so switch to `virtualenv` instead
- `python setup.py develop` is now even more heavily discouraged

### DIFF
--- a/.github/workflows/python-package.yml
+++ b/.github/workflows/python-package.yml
@@ -14,7 +14,7 @@ jobs:
         os: [ubuntu-latest, macos-latest]
         # Use only lowest and highest supported python versions for now,
         # to speed up CI runs
-        python-version: [3.8, "3.11"]
+        python-version: [3.8, "3.12"]
         node-version: [16]
       fail-fast: false
 
--- a/.github/workflows/python-package.yml
+++ b/.github/workflows/python-package.yml
@@ -47,6 +47,7 @@ jobs:
     - name: Generate Isso package
       id: generate-package
       run: |
+        pip install setuptools
         python setup.py sdist
         echo "::set-output name=package_file::$(ls dist/)"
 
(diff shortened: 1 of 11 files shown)


jelmer:
FWIW these changes LGTM, apart from the merge conflicts :)

ix5:
Might also be time to move on from `python setup.py {sdist, bdist}` and use the "official" PyPA [build](https://pypi.org/project/build/) package instead, although that's _another_ dependency. PyPA [recommends it](https://packaging.python.org/en/latest/tutorials/packaging-projects/#generating-distribution-archives) (via [stackoverflow](https://stackoverflow.com/questions/62604916/how-to-build-an-sdist-with-pip/75393059#75393059))

PR #1019, opened 2024-04-27
db: Make 'text' field in 'comments' table NOT NULL and handling data migration

pkvach:
## Checklist
      

This update introduces a schema migration to version 4 for the database, focusing on enhancing the 'comments' table. This ensures that the 'text' field in the 'comments' table will always have a value, which improves data consistency and integrity.

The key changes:
- Incrementing the database schema version from 3 to 4.
- Updating the "text" field in the "comments" table to be NOT NULL by setting empty strings for any NULL values.
- Creating a new table, "comments_new", with a NOT NULL constraint on the "text" field and then migrating data from the old "comments" table to this new table.
- Renaming the original "comments" table to "comments_backup_v3" for backup purposes and renaming "comments_new" to "comments" to finalize the migration.
- The migration process includes transaction handling to ensure data integrity, with a rollback in case of any errors during the migration.

## Why is this necessary?

- https://github.com/isso-comments/isso/issues/979
- https://github.com/isso-comments/isso/pull/994

### DIFF
--- a/isso/db/__init__.py
+++ b/isso/db/__init__.py
@@ -22,7 +22,7 @@ class SQLite3:
     a trigger for automated orphan removal.
     """
 
-    MAX_VERSION = 3
+    MAX_VERSION = 5
 
     def __init__(self, path, conf):
 
@@ -68,7 +68,7 @@ class SQLite3:
         if self.version >= to:
             return
 
-        logger.info("migrate database from version %i to %i", self.version, to)
+        logger.info("Migrating database from version %i to %i", self.version, to)
 
         # re-initialize voters blob due a bug in the bloomfilter signature
         # which added older commenter's ip addresses to the current voters blob
@@ -78,20 +78,36 @@ class SQLite3:
             bf = memoryview(Bloomfilter(iterable=["127.0.0.0"]).array)
 
             with sqlite3.connect(self.path) as con:
-                con.execute('UPDATE comments SET voters=?', (bf, ))
-                con.execute('PRAGMA user_version = 1')
-                logger.info("%i rows changed", con.total_changes)
+                con.execute("BEGIN TRANSACTION")
+                try:
+                    con.execute('UPDATE comments SET voters=?', (bf, ))
+                    con.execute('PRAGMA user_version = 1')
+                    con.execute("COMMIT")
+                    logger.info("Migrating DB version 0 to 1 by re-initializing voters blob, %i rows changed",
+                                con.total_changes)
+                except sqlite3.Error as e:
+                    con.execute("ROLLBACK")
+                    logger.error("Migrating DB version 0 to 1 failed: %s", e)
+                    raise RuntimeError("Migrating DB version 0 to 1 failed: %s" % e)
 
         # move [general] session-key to database
         if self.version == 1:
 
             with sqlite3.connect(self.path) as con:
-                if self.conf.has_option("general", "session-key"):
-                    con.execute('UPDATE preferences SET value=? WHERE key=?', (
-                        self.conf.get("general", "session-key"), "session-key"))
+                con.execute("BEGIN TRANSACTION")
+                try:
+                    if self.conf.has_option("general", "session-key"):
+                        con.execute('UPDATE preferences SET value=? WHERE key=?', (
+                            self.conf.get("general", "session-key"), "session-key"))
 
-                con.execute('PRAGMA user_version = 2')
-                logger.info("%i rows changed", con.total_changes)
+                    con.execute('PRAGMA user_version = 2')
+                    con.execute("COMMIT")
+                    logger.info("Migrating DB version 1 to 2 by moving session-key to database, %i rows changed",
+                                con.total_changes)
+                except sqlite3.Error as e:
+                    con.execute("ROLLBACK")
+                    logger.error("Migrating DB version 1 to 2 failed: %s", e)
+                    raise RuntimeError("Migrating DB version 1 to 2 failed: %s" % e)
 
         # limit max. nesting level to 1
         if self.version == 2:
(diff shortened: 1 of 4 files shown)


isso/db/__init__.py
```diff
@@ -121,3 +121,39 @@ def first(rv):
 
                 con.execute('PRAGMA user_version = 3')
                 logger.info("%i rows changed", con.total_changes)
+
+        # "text" field in "comments" became NOT NULL
+        if self.version == 3:
+            with sqlite3.connect(self.path) as con:
+                con.execute("BEGIN TRANSACTION")
```
ix5, line 128: I like this wrapping in a `TRANSACTION`, it's good practice. Might make sense to do this for the other migration steps as well.

isso/db/__init__.py
```diff
@@ -121,3 +121,39 @@ def first(rv):
...
+                con.execute("UPDATE comments SET text = '' WHERE text IS NULL")
+                logger.info("%i rows changed", con.total_changes)
```
ix5, line 130: I know this is a copy-paste from the other migration steps, but can we make this more informative? E.g. `Migrating DB version 3 to 4 by setting empty comments.text to '', %i rows changed`

isso/db/__init__.py
```diff
@@ -121,3 +121,39 @@ def first(rv):
...
+                        INSERT INTO comments_new (
+                            tid, id, parent, created, modified, mode, remote_addr, text, author, email, website, likes, dislikes, voters, notification
+                        )
+                        SELECT
+                            tid, id, parent, created, modified, mode, remote_addr, text, author, email, website, likes, dislikes, voters, notification
+                        FROM comments
+                    """)
+
+                    # swap tables
+                    con.execute("ALTER TABLE comments RENAME TO comments_backup_v3")
```
ix5, line 152: I personally don't like littering our users' database with unused tables.

This might also "break" promises to users' setups where they scrub email/IP PII from `comments` or vacuum comments older than x days. They would not expect this info to be retained and this might even create legal troubles for them.

We should rather be confident that this migration succeeds and does not destroy data. You have written tests and been cautious. If the migration fails, you auto-rollback. We can `DROP` the backup table imo.

isso/db/__init__.py
```diff
@@ -121,3 +121,39 @@ def first(rv):
...
+
+        # "text" field in "comments" became NOT NULL
+        if self.version == 3:
+            with sqlite3.connect(self.path) as con:
+                con.execute("BEGIN TRANSACTION")
+                con.execute("UPDATE comments SET text = '' WHERE text IS NULL")
+                logger.info("%i rows changed", con.total_changes)
+
+                # create new table with NOT NULL constraint for "text" field
+                con.execute('CREATE TABLE IF NOT EXISTS comments_new ('
```
ix5, line 133: This must be kept in sync with [db/comments.py](https://github.com/isso-comments/isso/blob/27bdc72a5e2f4e6ae0a9fd506225774ae4ed711f/isso/db/comments.py). Could you add a note there and also here cautioning authors of this pitfall?
Or even better, refactor a bit so that this executes `db.comments.__init__()` on `comments_new`?

isso/tests/test_db.py
```diff
@@ -118,3 +118,58 @@ def test_limit_nested_comments(self):
...
+            con.execute(
+                "INSERT INTO threads (uri, title) VALUES (?, ?)", ("/", "Test"))
+
+            con.execute("INSERT INTO comments (id, parent, created, text, voters) VALUES (?, ?, ?, ?, ?)",
+                        (1, None, 1, None, sqlite3.Binary(b"")))
+
+            con.execute("INSERT INTO comments (id, parent, created, text, voters) VALUES (?, ?, ?, ?, ?)",
+                        (2, 1, 2, "foo", sqlite3.Binary(b"")))
+
+        conf = config.new({
```
ix5, line 158: duplicate, you already define this in line 123

isso/tests/test_db.py
```diff
@@ -118,3 +118,58 @@ def test_limit_nested_comments(self):
...
+            }
+        })
+
+        db = SQLite3(self.path, conf)
+
+        self.assertEqual(db.version, SQLite3.MAX_VERSION)
+
+        with sqlite3.connect(self.path) as con:
+            # assert that the "text" field has "NOT NULL" constraint
+            rv = con.execute("SELECT \"notnull\" FROM pragma_table_info('comments') WHERE name='text'").fetchone()
```
ix5, line 171: nit, but why `\"notnull\"` and not single-quote `'notnull'`?

ix5 (commented):
Nice effort, thank you.

I noticed that these lines should have been a migration as well: https://github.com/isso-comments/isso/blob/27bdc72a5e2f4e6ae0a9fd506225774ae4ed711f/isso/db/comments.py#L44-47

Could you add them while you're at it?

isso/tests/test_db.py
pkvach: In single quotes `'notnull'` is treated as a string.

pkvach:
> Nice effort, thank you.
> 
> I noticed that these lines should have been a migration as well:
> 
> https://github.com/isso-comments/isso/blob/27bdc72a5e2f4e6ae0a9fd506225774ae4ed711f/isso/db/comments.py#L44-L47
> 
> Could you add them while you're at it?

Thanks for the review.
I have added the suggested changes. I will do squash commits once the code review is complete.

isso/db/__init__.py
```diff
@@ -78,20 +78,32 @@ def migrate(self, to):
...
-                con.execute('PRAGMA user_version = 1')
-                logger.info("%i rows changed", con.total_changes)
+                con.execute("BEGIN TRANSACTION")
+                try:
+                    con.execute('UPDATE comments SET voters=?', (bf, ))
+                    con.execute('PRAGMA user_version = 1')
+                    con.execute("COMMIT")
+                    logger.info("Database migration to version 1 finished, %i rows changed", con.total_changes)
+                except sqlite3.Error as e:
+                    logger.error("Database migration to version 1 failed: %s", e)
```
ix5, line 88: If a migration fails (which should _never_ happen except for corrupted or fiddled-with databases), should we bail (`sys.exit(1)` / `raise`) out entirely? `isso.db` will otherwise attempt to write to non-existent DB columns and fail, potentially losing comments.

isso/db/__init__.py
```diff
@@ -114,10 +126,72 @@ def first(rv):
...
+                    con.execute("COMMIT")
+                    logger.info("Database migration to version 3 finished, %i rows changed", con.total_changes)
+                except sqlite3.Error as e:
+                    logger.error("Database migration to version 3 failed: %s", e)
+                    con.execute("ROLLBACK")
+
+        # add "notification" field to "comments" (moved from Comments class to migration)
+        if self.version == 3:
+            with sqlite3.connect(self.path) as con:
+                self.migrate_to_version_4(con)
```
ix5, line 146: Is there a particular reason you decided to factor out only this migration step into its own function? Afaics you're only saving one indentation level.

isso/db/__init__.py
```diff
@@ -114,10 +126,72 @@ def first(rv):
...
+
+        # "text" field in "comments" became NOT NULL
+        if self.version == 4:
+            with sqlite3.connect(self.path) as con:
+                con.execute("BEGIN TRANSACTION")
+                con.execute("UPDATE comments SET text = '' WHERE text IS NULL")
+                logger.info("Migrating DB version 4 to 5 by setting empty comments.text to '', %i rows changed",
```
ix5, line 153: Can you please keep the log messages consistent?

Currently a full migration looks like this:
```
migrate database from version 0 to 5
Database migration to version 1 finished, 0 rows changed
Database migration to version 2 finished, 0 rows changed
Database migration to version 3 finished, 0 rows changed
Database migration to version 4 skipped, column 'notification' already exists
Migrating DB version 4 to 5 by setting empty comments.text to '', 0 rows changed
```

Better to emit this log at the end of the migration (after the `COMMIT`):
`Database migration to version 5 finished, %i rows changed by setting empty comments.text to ''`

Also, suggestion for line 71:
```diff
-        logger.info("migrate database from version %i to %i", self.version, to)
+        logger.info("Migrating database from version %i to %i", self.version, to)
```

isso/db/comments.py
```diff
@@ -31,20 +31,33 @@ class Comments:
               'remote_addr', 'text', 'author', 'email', 'website',
               'likes', 'dislikes', 'voters', 'notification']
 
+    # ! It's being used in migration from version 4 to 5
```
ix5, line 34:
```suggestion
    # This method is used in the migration script from version 4 to 5. You need to write a new migration if you change the database schema!
```

isso/tests/test_db.py
ix5: Ah, thank you.

ix5 (commented):
This is a massive improvement, thank you! A few suggestions inline, feel free to ignore if you feel they are too nitpicky.

isso/db/__init__.py
pkvach: I thought it would be more readable that way. In order not to create additional nesting due to checking that the `notification` field already exists.

isso/db/__init__.py
pkvach: That's a good point, thank you.
I've added an exception for such cases.

pkvach:
@ix5 
Thank you for your suggestions. I've made the changes.

ix5 (approved):

isso/db/__init__.py
ix5: There is too much nesting, I agree.

isso/db/__init__.py
ix5: Great, thanks.

Just FYI, emitting a `logger.error` and then throwing a `RuntimeError` with the same message is a bit superfluous, I'll fix that in a follow-up commit.

ix5:
Squashed (pushed to your branch) and merged.

I have a few changes that I wrote and stashed locally when reviewing this, but that's for another day. This PR is a great improvement that allows for more confidence when releasing the next version.

pkvach:
Great, thank you.

PR #1020, opened 2024-04-27
admin: Add support for comment search by Thread URL in admin interface

pkvach:
## Checklist
      

The main theme of the changes is to enhance the comment search functionality in the admin interface. The changes allow searching for comments not only by comment URL but also by thread URL.

## Why is this necessary?

- https://github.com/isso-comments/isso/issues/642
- https://github.com/isso-comments/isso/pull/1000

### DIFF
--- a/isso/css/admin.css
+++ b/isso/css/admin.css
@@ -176,3 +176,70 @@ a {
     width: 100%;
     margin-top: 1em;
 }
+
+
+.search-tooltip {
+  position: relative;
+  display: inline-block;
+  cursor: pointer;
+  user-select: none;
+}
+
+.search-tooltip-icon {
+  display: inline-block;
+  width: 1.2em;
+  height: 1.2em;
+  border-radius: 50%;
+  background: #fffaf3;
+  font-family: serif;
+  font-weight: bold;
+  text-align: center;
+  color: #000;
+  line-height: 1.1em;
+  border: 1px solid rgba(0, 0, 0, 0.2);
+  box-shadow: 0 1px 2px rgba(0, 0, 0, 0.1);
+}
+
+.search-tooltip .search-tooltip-text {
+  visibility: hidden;
+  width: 370px;
+  background-color: #555;
+  color: #fff;
+  text-align: left;
+  border-radius: 6px;
+  padding: 8px;
+  position: absolute;
+  z-index: 1;
+  bottom: 125%;
+  left: 50%;
+  margin-left: -185px;
+}
+
+/* Tooltip arrow */
+.search-tooltip .search-tooltip-text::after {
+  content: "";
+  position: absolute;
+  top: 100%;
+  left: 50%;
+  margin-left: -5px;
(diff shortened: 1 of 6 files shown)


ix5 (changes requested):
Thanks for the PR!

Could we perhaps find a better way to convey to admins that they can enter both thread and individual comment URLs?

Also, would it make sense to make the search field a bit smarter and just let it accept plain numbers (for comment ids) as well?

pkvach:
> Thanks for the PR!

I'm happy to help :)

> Could we perhaps find a better way to convey to admins that they can enter both thread and individual comment URLs?

There is an option to add a tooltip with explanatory text. This will not take up too much space.

> Also, would it make sense to make the search field a bit smarter and just let it accept plain numbers (for comment ids) as well?

Then we will definitely need explanatory text and will have to give up `<input type=“url”>`.

ix5:
> > Could we perhaps find a better way to convey to admins that they can enter both thread and individual comment URLs?
> 
> There is an option to add a tooltip with explanatory text. This will not take up too much space.

Good idea. Something like `<abbr>` that pops on mouseover will not work on mobile, but if you have a good solution, go for it.

> > Also, would it make sense to make the search field a bit smarter and just let it accept plain numbers (for comment ids) as well?
> 
> Then we will definitely need explanatory text and will have to give up `<input type=“url”>`.

On second thought, let's not implement the "smart" search and leave it as is. It would just be confusing.

pkvach:
I added a tooltip with help text next to the search field.
It's ready for review.

isso/css/admin.css
```diff
@@ -176,3 +176,67 @@ a {
...
+
+.search-tooltip .show {
+  visibility: visible;
+  -webkit-animation: fadeIn 1s;
+  animation: fadeIn 1s;
+}
+
+/* Add animation (fade in the tooltip) */
+@-webkit-keyframes fadeIn {
+  from {opacity: 0;}
```
ix5, line 235: single trailing whitespace here
```diff
-  from {opacity: 0;} 
+  from {opacity: 0;}
```

ix5 (commented):
Nice, the tooltip is really useful.

I'd just want it a bit more integrated/look alike the rest of the surrounding buttons, so that it's clearer it is clickable.

Current:
Screenshot from 2024-05-03 19-03-18

With more prominent border and shadow:
Screenshot from 2024-05-03 19-02-55

pkvach:
> Nice, the tooltip is really useful.
> 
> I'd just want it a bit more integrated/look alike the rest of the surrounding buttons, so that it's clearer it is clickable.
> 
> Current: Screenshot from 2024-05-03 19-03-18
> 
> With more prominent border and shadow: Screenshot from 2024-05-03 19-02-55

Thank you for the review and a nice addition.
I made the suggested changes and did squash commits.

ix5 (approved):

ix5:
Merged, thanks!

PR #1028, opened 2024-05-31
API Access to Moderation Queue

gflohr:
## Checklist
      

This PR introduces a new API endpoint `/pending` which works like `/latest` with the following differences:

* Requests must be authorized with HTTP basic authentication using username "admin" and the admin password.
* The endpoint is enabled in the configuration with `[general] pending-enabled` set to "true".
* The endpoint only returns posts in the moderation queue.
* The endpoint additionally requires the `admin` interface to be enabled.

The last requirement is maybe a little bit paranoid. Rationale: the admin password is currently only used for the admin interface. The requirement is meant to ensure that people have changed the default password. On the other hand, the endpoint does not expose sensitive information but only potential spam.

The feature/endpoint only makes sense, when moderation is enabled.  But this is not explicitly checked because it causes no harm if moderation is *not* enabled.

## Why is this necessary?
If you want to implement an alternative notification mechanism for comments awaiting moderation, access to the moderation queue via the API is required. The existing `/latest` endpoint only returns accepted comments.  The new endpoint should require authorization because it exposes comments that are not publicly visible.
...

### DIFF
--- a/contrib/isso-dev.cfg
+++ b/contrib/isso-dev.cfg
@@ -18,6 +18,7 @@ notify = stdout
 reply-notifications = false
 log-file =
 latest-enabled = true
+pending-enabled = true
 
 [admin]
 enabled = true
--- a/isso/isso.cfg
+++ b/isso/isso.cfg
@@ -65,6 +65,10 @@ gravatar-url = https://www.gravatar.com/avatar/{}?d=identicon&s=55
 # needing to previously know the posts URIs)
 latest-enabled = false
 
+# enable the "/pending" endpoint, that works likes "/latest" but only
+# for comments waiting moderation
+pending-enabled = false
+
 [admin]
 enabled = false
 
--- a/isso/tests/test_comments.py
+++ b/isso/tests/test_comments.py
@@ -5,10 +5,12 @@ import os
 import re
 import tempfile
 import unittest
+import base64
 
 from urllib.parse import urlencode
 
 from werkzeug.wrappers import Response
+from werkzeug.datastructures import Headers
 
 from isso import Isso, core, config
 from isso.utils import http
@@ -689,6 +691,20 @@ class TestComments(unittest.TestCase):
         response = self.get('/latest?limit=5')
         self.assertEqual(response.status_code, 404)
 
+    def testPendingNotFound(self):
+        # load some comments in a mix of posts
+        saved = []
+        for idx, post_id in enumerate([1, 2, 2, 1, 2, 1, 3, 1, 4, 2, 3, 4, 1, 2]):
+            text = 'text-{}'.format(idx)
+            post_uri = 'test-{}'.format(post_id)
+            self.post('/new?uri=' + post_uri, data=json.dumps({'text': text}))
+            saved.append((post_uri, text))
+
+        response = self.get('/pending?limit=5')
+
+        # If the admin interface was not enabled we should get a 404.
+        self.assertEqual(response.status_code, 404)
+
 
 class TestHostDependent(unittest.TestCase):
 
@@ -763,6 +779,8 @@ class TestModeratedComments(unittest.TestCase):
         conf.set("moderation", "enabled", "true")
         conf.set("guard", "enabled", "off")
         conf.set("hash", "algorithm", "none")
+        conf.set("admin", "enabled", "true")
+        self.conf = conf
 
         class App(Isso, core.Mixin):
             pass
@@ -770,6 +788,8 @@ class TestModeratedComments(unittest.TestCase):
         self.app = App(conf)
         self.app.wsgi_app = FakeIP(self.app.wsgi_app, "192.168.1.1")
         self.client = JSONClient(self.app, Response)
+        self.post = self.client.post
+        self.get = self.client.get
 
     def tearDown(self):
         os.unlink(self.path)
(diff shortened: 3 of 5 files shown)


isso/tests/test_comments.py
```diff
@@ -844,6 +864,64 @@ def testModerateComment(self):
...
+        return self.client.get(url, headers=headers)
+
+    def testPendingNotEnabled(self):
+        password = "s3cr3t"
+        self.conf.set("admin", "enabled", "true")
+        self.conf.set("admin", "password", password)
+        response = self.getAuthenticated('/pending?limit=5', 'admin', password)
+        self.assertEqual(response.status_code, 404)
+
+    def testPendingNotEnabled(self):
```
ix5, line 891: should be `testPendingEnabled`

isso/views/comments.py
```diff
@@ -1562,6 +1590,77 @@ def latest(self, environ, request):
...
+    @requires_admin
+    @requires_auth
+    def pending(self, environ, request):
+        # if the feature is not allowed, don't present the endpoint
+        if not self.conf.getboolean("pending-enabled"):
+            return NotFound(
+                "Unavailable because 'pending-enabled' not set by site admin"
+            )
+
+        return self._latest(environ, request, "2")
```
ix5, line 1662: Rather than adding a new endpoint, wouldn't it make more sense to instead pass through a `mode` GET parameter? It would remain at `"1"` as a default for `/latest` and could be set to `"2"` to get pending comments (which would then require auth).

isso/isso.cfg
```diff
@@ -65,6 +65,10 @@ gravatar-url = https://www.gravatar.com/avatar/{}?d=identicon&s=55
 # needing to previously know the posts URIs)
 latest-enabled = false
 
+# enable the "/pending" endpoint, that works likes "/latest" but only
+# for comments waiting moderation
```
ix5, line 69: "that works like"
"awaiting"

isso/tests/test_comments.py
```diff
@@ -844,6 +864,64 @@ def testModerateComment(self):
         # Comment should no longer exist
         self.assertEqual(self.app.db.comments.get(id_), None)
 
+    def testPendingWithoutAdmin(self):
+        self.conf.set("admin", "enabled", "false")
```
ix5, line 868: Are these tests guaranteed to be run in order? I'd prefer you un-set the `admin.enabled` value at test end.

ix5 (commented):
In general, I think this idea is sensible. A few comments inline.

I don't think you really need a new config option for this though. Either someone has already enabled the admin interface, giving anyone with admin credentials access to *all* comments, or admin is disabled, which will also disable the new `/pending` endpoint. My concern is that going at this rate, the whole routing system will at some point in the future be (poorly) exposed through the configuration file.

gflohr:
I'm afraid, I will need a couple of weeks to come back to this PR.

jelmer:
Hi @gflohr , any news on this?

gflohr:
I'll try to find time within the next couple of days.

isso/isso.cfg
jelmer: This is still relevant

jelmer (changes requested):

contrib/isso-dev.cfg
```diff
@@ -18,6 +18,7 @@ notify = stdout
 reply-notifications = false
 log-file =
 latest-enabled = true
+pending-enabled = true
```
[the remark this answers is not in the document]
gflohr: The endpoint and the configuration value have been deleted. You now have to use a parameter `mode` for the `/latest` endpoint instead.

isso/tests/test_comments.py
gflohr: done

isso/views/comments.py
gflohr: done

contrib/isso-dev.cfg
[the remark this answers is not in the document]
jelmer: The same question applies for the latest endpoint though - do we need a configuration setting?

isso/views/comments.py
```diff
@@ -1518,12 +1545,19 @@ def admin(self, env, req):
...
     @apiVersion 0.12.6
     @apiDescription
-        Get the latest comments from the system, no matter which thread. Only available if `[general] latest-enabled` is set to `true` in server config.
+        Get the latest accepted comments from the system, no matter which thread. Only available if `[general] latest-enabled` is set to `true` in server config.
 
     @apiQuery {Number} limit
         The quantity of last comments to retrieve
 
-    @apiExample {curl} Get the latest 5 comments
+    @apiQuery {Number{1,2}} [mode=1]
```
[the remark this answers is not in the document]
gflohr: https://github.com/isso-comments/isso/pull/1028#discussion_r1660209344 suggested the numeric constants. But make a suggestion for string constants.

isso/tests/test_comments.py
```diff
@@ -860,6 +911,38 @@ def testModerateComment(self):
...
+        for idx, post_id in enumerate([1, 2, 2, 1, 2, 1, 3, 1, 4, 2, 3, 4, 1, 2]):
+            text = 'text-{}'.format(idx)
+            post_uri = 'test-{}'.format(post_id)
+            self.post('/new?uri=' + post_uri, data=json.dumps({'text': text}))
+            saved.append((post_uri, text))
+
+        password = "s3cr3t"
+        self.conf.set("admin", "enabled", "true")
+        self.conf.set("admin", "password", password)
+        self.conf.set("general", "latest-enabled", "true")
```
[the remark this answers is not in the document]
gflohr: I think that suggestion is outside of the scope of this PR.

isso/views/comments.py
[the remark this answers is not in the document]
jelmer: mode=pending vs mode=approved ?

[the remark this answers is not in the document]
gflohr: Okay. I will change that then.

PR #1031, opened 2024-06-17
Invert screenshot for dark color theme

oberrich:
## Checklist
- [x] Did none of the above
This inverts the screenshots on the README of this repo.

## Why is this necessary?
My eyes, bug prevention

For comparison see [`patch-1`](https://github.com/oberrich/isso/tree/patch-1?tab=readme-ov-file#screenshot) vs. [`master`](https://github.com/oberrich/isso?tab=readme-ov-file#screenshot). You may want to change the URL of the dark mode logo so it doesn't go away when I delete the `patch-1` tag.

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,7 +22,7 @@ details and [documentation](https://isso-comments.de/docs/).
 ## Screenshot
 
 <picture>
- <source media="(prefers-color-scheme: dark)" srcset="https://github.com/isso-comments/isso/assets/6305520/741b312e-798c-412b-918d-39d96deb4a5d">
+ <source media="(prefers-color-scheme: dark)" srcset="https://github.com/isso-comments/isso/assets/6305520/39892cd1-2930-4b73-ba8b-88ec6d1da4ce">
  <img alt="Isso in Action" src="https://user-images.githubusercontent.com/10212877/167268553-3f30b448-25ff-4850-afef-df2f2e599c93.png">
 </picture>
 
--- a/README.md
+++ b/README.md
@@ -21,7 +21,10 @@ details and [documentation](https://isso-comments.de/docs/).
 
 ## Screenshot
 
-![Isso in Action](https://user-images.githubusercontent.com/10212877/167268553-3f30b448-25ff-4850-afef-df2f2e599c93.png)
+<picture>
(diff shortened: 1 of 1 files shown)


jelmer:
Please include the assets in the repo, rather than outside of it.

I guess doing this adds to the burden of updating the screenshot if we need to refresh it in the future. Not a big deal, but worth pointing out.

oberrich:
I don't think the screenshot is hosted inside the repo, if I copy & paste the image directly into my clone of this repository it uploads it to `github.com/oberrich`, so I guess the best place I can host this image for you is right here:
image

ix5:
The screenshot not being hosted inside the repo is due to me updating it from an external source (posativ.org) to one uploaded to the corresponding PR: https://github.com/isso-comments/isso/pull/867

That wasn't the correct way to do it, we should add the light (and potentially dark mode) screenshots to this repo and link them from inside the markdown. I think GitHub's markdown parser can resolve (relative) links to files inside the _current_ repo so that we don't have to hardcode the path to `user-images.githubusercontent.com` (example: `Isso in action`). That way, forks updating the screenshot files inside their repo don't have to wonder why the README image doesn't update.

Then again, if we want the images displayed correctly on PyPI, we need hardcoded paths, no?
Probably something like `https://raw.githubusercontent.com/isso-comments/isso/master/isso/img/isso.svg` is what we want, then.

As for re-generating the screenshots: I've tried to document that a bit inside https://github.com/isso-comments/isso-marketing-material but maybe someone can submit a PR to add a `Dockerfile` to automate that step.

oberrich:
`isso/img/` and hard coded URI is probably most reasonable and portable.

ix5:
> `isso/img/` and absolute URL is probably most reasonable and portable.

Sounds good. Please add a commit that places both images (light and dark) there and update the README (in another commit) to reference them.

Would also be nice to have a quickly reproducible way to (automatically) regenerate the screenshots.

oberrich:
> > `isso/img/` and absolute URL is probably most reasonable and portable.
> 
> Sounds good. Please add a commit that places both images (light and dark) there and update the README (in another commit) to reference them.
> 
> Would also be nice to have a quickly reproducible way to (automatically) regenerate the screenshots.

Feel free to commit it on your own, I've decided not to spend any more time on this issue.

PR #1035, opened 2024-08-24
server: Set reply sorting to 'oldest'

### LINKED ISSUE #1026: [Feature Request] Separate sorting for replies
## Checklist
- [x] I am running the latest version. Installing Isso from GitHub from the `master` branch does not fix my issue
- [x] I have searched the [open issues](https://github.com/isso-comments/isso/issues), but my issue has not already been reported

## What is not working?
When using the new sorting system introduced in #1005, replies are sorted the exact same as regular comments. If comments are sorted by newest first, the replies to said comments will also be sorted as such. This can potentially confuse the reader, as the reading order for threads is usually always oldest first. [Here's an example of what it looks like now:](https://www.ggtyler.dev/other/frontends#isso-335)
An example comment thread
Note how the thread starts with the oldest post (the OP), then immediately after that, the newest post (the reply.)

## How can one reproduce this issue?
On the client-side configuration, set `data-isso-sorting` to `newest`. *(This is possibly also visible in `upvotes` too, but haven't tested yet.)* Then, post a comment, and reply to it at least twice.

ggtylerr:
## Checklist
      

Quick SQL patch to fix #1026 

## Why is this necessary?
Current reply sorting leads to confusion. See linked issue for details.

NOTE: I tested this briefly and it seems to be working (and of course unit tests pass) but I wasn't sure if the sorting option was actually going through. Hard to tell on the demo page. Will probably test tmrw over curl.


isso/db/comments.py
```diff
@@ -217,16 +217,16 @@ def fetchall(self, mode=5, after=0, parent='any', order_by='id',
 
         # custom sanitization
         if order_by not in ['id', 'created', 'modified', 'likes', 'dislikes', 'tid']:
-            sql.append('ORDER BY ')
-            sql.append("comments.created")
+            sql.append('ORDER BY CASE WHEN comments.parent IS NOT NULL THEN comments.id END, ')
```
ix5, line 220: This first `if` case is already the fallback for when no sorting is specified. No need to alter anything here, instead leave `comments.created` forsorting.

isso/db/comments.py
```diff
@@ -217,16 +217,16 @@ def fetchall(self, mode=5, after=0, parent='any', order_by='id',
...
+            sql.append('comments.created')
             if not asc:
                 sql.append(' DESC')
         else:
-            sql.append('ORDER BY ')
+            sql.append('ORDER BY CASE WHEN comments.parent IS NOT NULL THEN comments.id END, ')
```
ix5, line 225: Please use `comments.created` instead of `comments.id` for sorting. The `id` field is the (auto-incrementing) primary key. A PK id might be reassigned when a comment is deleted and a subsequently created comment grabs the next available id, shuffling the sort order.

isso/db/comments.py
```diff
@@ -258,7 +258,7 @@ def fetch(self, uri, mode=5, after=0, parent='any',
         # custom sanitization
         if order_by not in ['id', 'created', 'modified', 'likes', 'dislikes', 'karma']:
             order_by = 'id'
```
ix5, line 260: I really don't like this fallback to sort by `id` instead of `created`. Let's fix it here as well.

isso/db/comments.py
```diff
@@ -258,7 +258,7 @@ def fetch(self, uri, mode=5, after=0, parent='any',
...
-        sql.append('ORDER BY ')
+        sql.append('ORDER BY CASE WHEN comments.parent IS NOT NULL THEN comments.id END, ')
```
ix5, line 261: Will a `CASE WHEN .. END` immediately terminate the SQL statement, skipping over the remaining supplied arguments (`LIMIT`, ...)?
I'm not sure, could you please check?

ix5 (commented):
Hi, thank you for the quick patch! Left a few inline comments, mainly regarding the usage of `comments.id` for sorting which I'd prefer you'd switch to `comments.created` for consistency.

isso/db/comments.py
ggtylerr: Sure thing, I'll switch to it and make another commit when I get off of work in a few hours.

isso/db/comments.py
ggtylerr: It shouldn't, AFAIK it just functions like the pre-existing order by, so no termination. But I'll double check it just in case.

isso/db/comments.py
ggtylerr: Switched to it (sorry this took so long, had some server nonsense pop up.

isso/db/comments.py
ggtylerr: Does not appear to be the case:
image
image
(Demo page above is shown with sorting set to oldest.)

ggtylerr:
Just confirmed the patch to be working now - you can see the results on my site. Here's the comment I showed as an example last time:
image

ix5:
Thank you for addressing the feedback. Please squash your commits and give the resulting commit a more descriptive title/body, such as `db: Set reply sorting to 'oldest'`.

You also checked in an updated `package-lock.json` to your commit when you didn't mean to, please remove that as well.

ggtylerr:
Fixed up as requested

Kerumen:
@ix5 do you plan to merge and release this? Thanks 🙏🏻

ix5:
Sorry for leaving this in limbo for so long.
The test suite (predictably) fails for me:
```
FAILED isso/tests/test_comments.py::TestComments::testFeed - AssertionError: '"1-1"' != '"1-2"'
FAILED isso/tests/test_comments.py::TestComments::testGetSortedByNewestWithNested - AssertionError: Lists differ: [2, 3, 4, 5, 6] != [6, 5, 4, 3, 2]
FAILED isso/tests/test_comments.py::TestComments::testGetSortedByUpvotesWithNested - AssertionError: Lists differ: [2, 3, 4, 5, 6] != [6, 3, 2, 4, 5]
```

@ggtylerr would you mind updating the tests to reflect the new sorting? Thank you!

jelmer:
Hi @ggtylerr any news on this?

ggtylerr:
Hi, sorry, I didn't get any notification for this when ix5 replied for some reason. I'll update it whenever I'm off my shift today.

ggtylerr:
I fetched upstream, cleared the merge conflict and updated the tests. Everything seems to work when I run it with nektos act, although I can't run isso itself since pip isn't letting me pull misaka.

One thing to note is that `testGetSortedByNewestWithNested` and `testGetSortedByUpvotesWithNested` now have the same order, due to them testing for the order of replies and not newest/upvotes. I consider this fine and these two still test whether or not the reply order is modified, but it's still something to note. Might need to modify to test multiple comments + replies instead of multiple comments, or just have a separate test.

(Also I cannot for the life of me figure out how to squash nicely w/ the merge commit, I'm sorry :cry:)

jelmer:
LGTM. I'll wait a few days to give @ix5 a chance to object. I'll merge otherwise.

Kerumen:
@jelmer Do you plan to cut a release including this change?

jelmer:
We should do another release at some point; first we should probably fix the failing tests though.

PR #1036, opened 2024-09-16
Fixing error message: URI {uri} does not exist

Lucas-C:
## Checklist
- [NA] (If adding features:) I have added tests to cover my changes
- [NA] (If docs changes needed:) I have updated the **documentation** accordingly.
- [NA] I have added an entry to `CHANGES.rst` because this is a user-facing change or an important bugfix

Previous to this one-line fix, the following error message could be returned:

    URI does not exist %s

Now the error message containes the actual URI:

    URI {uri} does not exist

## Why is this necessary?
To help debug problems! 🙂

### DIFF
--- a/isso/views/comments.py
+++ b/isso/views/comments.py
@@ -350,7 +350,7 @@ class API(object):
                         if resp and resp.status == 200:
                             uri, title = parse.thread(resp.read(), id=uri)
                         else:
-                            return NotFound('URI does not exist %s')
+                            return NotFound(f'URI {uri} does not exist')
                 else:
                     title = data['title']
 


jelmer:
The change overall LGTM, thanks! Do translations play well with f-strings?

jelmer:
What's going on with the failing e2e test?

Lucas-C:
Seems like the CI pipeline fails due to "screenshots not matching".

But this PR should not have any visual impact.

Maybe the reference screenshots should be updated in a separate PR?
The process is detailed in this documentation section: https://github.com/isso-comments/isso/blob/master/docs/docs/technical-docs/testing-client.rst#screenshot-comparison

jelmer (approved):

simsinght:
Kind of unrelated, but I'm trying to debug an issue where I keep getting the older `URI does not exist %s`. I see its because I'm using the official docker image, and it hasn't been built for over 2 months so likely doesn't contain this change. 

What would trigger the docker image to rebuild?

BBaoVanC:
> Kind of unrelated, but I'm trying to debug an issue where I keep getting the older `URI does not exist %s`. I see its because I'm using the official docker image, and it hasn't been built for over 2 months so likely doesn't contain this change.
> 
> What would trigger the docker image to rebuild?

Are you using docker compose, and if so, either the `image:` or `build:`? Depending on which you're using, either `docker compose pull` or `docker compose build` should update it.

PR #1070, opened 2025-10-28
Avoid an error if `Offset.update()` hasn't been called yet.

### LINKED ISSUE #1069: Timing-dependent`embed.min.js` issue causes sporadic "Uncaught TypeError: Reduce" fatal errors
Using `isso` with a fast webserver such as Caddy (it's harder to reproduce this on Apache) can lead to the following error being triggered due to an empty array (the JavaScript runs too fast and finds an empty "Offset.prototype.update":

```
Uncaught TypeError: Reduce of empty array with no initial value
    at Array.reduce (<anonymous>)
    at t.localTime (embed.min.js:1:21682)
    at f (embed.min.js:1:27861)
    at Object.p [as insert] (embed.min.js:1:27930)
    at embed.min.js:1:60073
    at Array.forEach (<anonymous>)
    at embed.min.js:1:60051
    at embed.min.js:1:4902
```

The website then displays the correct `isso` comment box, but with no comments on the page, despite the Network tab showing them being properly requested and returned from the server by the API.

I started encountering this after upgrading my webserver from Apache to Caddy, constantly on Chrome, but sporadically on Firefox (one out of 3-4 loads), pointing to it being a timing issue and mainly happening on very fast browser + server interactions.

The solution proved to be quite simple -- check if the array is empty before calling `reduce()`, and just return the current time if it is:

```
--- isso/isso/js/embed.dev.js	2025-10-27 01:01:08.000000000 +0100
+++ isso/isso/js/embed.min.js	2025-10-27 01:33:20.843501806 +0100
@@ -686,6 +686,10 @@
 };

 Offset.prototype.localTime = function() {
+    if (this.values.length === 0) {
+        return new Date();
+    }
+
     return new Date((new Date()).getTime() - this.values.reduce(
         function(a, b) { return a + b; }) / this.values.length);
 };
```

This is based on the latest `master` git version, and tested to work fine under either Caddy or Apache, Chrome or Firefox (including mobile).

I am not very familiar with where this actually comes from in the JavaScript source (since the `embed...` files are all generated), but it should be possible to fix it there in a similar manner.

afranchuk:
Closes #1069.

## Checklist
      

This fixes a bug where `Offset.localTime()` will throw an error if called before `Offset.update()`.

## Why is this necessary?
#1069

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -58,6 +58,7 @@ Bugfixes & Improvements
 - Disable Postbox submit button on click, enable after response (`#993`_, pkvach)
 - Document title parameter and improve error handling for /new API (`#1058`_, pkvach)
 - Set reply sorting to always be oldest (`#1035`_, ggtylerr)
+- Fix ``Offset.localTime()`` failing if ``Offset.update()`` hasn't been called yet.
 
 .. _#951: https://github.com/posativ/isso/pull/951
 .. _#967: https://github.com/posativ/isso/pull/967
--- a/isso/js/app/globals.js
+++ b/isso/js/app/globals.js
@@ -10,7 +10,7 @@ Offset.prototype.update = function(remoteTime) {
 
 Offset.prototype.localTime = function() {
     return new Date((new Date()).getTime() - this.values.reduce(
-        function(a, b) { return a + b; }) / this.values.length);
+        function(a, b) { return a + b; }, 0) / this.values.length);
 };
 
 var offset = new Offset();


jelmer (approved):

afranchuk:
The E2E test failures seem unrelated to this change.

jelmer:
> The E2E test failures seem unrelated to this change.

Yeah, see https://github.com/isso-comments/isso/issues/1068

PR #1076, opened 2025-12-04
Add ntfy as a notification backend

yiyunliu:
## Checklist
- [ ] I have written **proper commit message(s)**
      

I wrote a [ntfy.sh](https://ntfy.sh/) backend based on what's in the Stdout subscriber object. Everything has been tested to work on with my ntfy setup, except for the new_thread callback, which i'm not sure exactly how to trigger.

`ntfy.sh` is quite easy to set up and satisfies my need to run the notification service locally on my VPN. SMTP, on the other hand, is kind of tricky to setup compared to the straightforward API that `ntfy` provides.

If the patch is considered useful, I'd be happy to start working on the checklist items. (Might need some help on the testing side)
## Why is this necessary?
N/A because it's a feature?

### DIFF
--- a/isso/ext/notifications.py
+++ b/isso/ext/notifications.py
@@ -5,12 +5,13 @@ import json
 import smtplib
 import socket
 import time
-import requests
 
 from _thread import start_new_thread
 from email.message import EmailMessage
 from email.utils import formatdate
 from urllib.parse import quote, urljoin
+from urllib.request import urlopen
+from urllib.error import URLError
 
 import logging
 logger = logging.getLogger("isso")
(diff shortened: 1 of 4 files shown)


isso/__init__.py
```diff
@@ -109,16 +109,22 @@ def __init__(self, conf):
...
             elif backend in ("smtp", "SMTP"):
                 smtp_backend = True
+            elif backend == "ntfy":
+                ntfy_backend = True
             else:
                 logger.warning("unknown notification backend '%s'", backend)
         if smtp_backend or conf.getboolean("general", "reply-notifications"):
             subscribers.append(SMTP(self))
 
+        if ntfy_backend:
```
ix5: No need to introduce a `ntfy_backend` variable, you can just move this block below `elif backend == "ntfy"`

isso/ext/notifications.py
```diff
@@ -210,6 +212,58 @@ def _retry(self, subject, body, to, headers):
...
+
+    def _activate_comment(self, thread, comment):
+        self.notify("comment %(id)s activated" % thread)
+
+    def notify(self, str):
+        topic_url = urljoin(self.conf.get("url"),self.conf.get("topic"))
+        try:
+            with urlopen(topic_url,str.encode(encoding='utf-8')) as response:
+                pass
+        except URLError as e:
```
ix5: Not that I think it matters much, but other types of errors (apart from `URLError`) are not caught and reported here.

ix5 (changes requested):
In case we decide to go along with this PR, this would be my suggested changes. @yiyunliu no need to act upon this yet before a decision has been made, just letting you know.

yiyunliu:
> @yiyunliu maybe you would also be up to the task since @ Guts seems to have stepped back a bit.

I'd be happy to take a look at the draft and try rebasing the PR to and resolving some of the remaining issues. I'm not really familiar with the spec of webhook, however, so I might have trouble coming up with good test cases other than checking if it works with ntfy's webhook support.

Issue #1040, opened 2024-10-12
Switch from misaka to something else?


fluffy-critter:
## Checklist
- [x] I am running the latest version. Installing Isso from GitHub from the `master` branch does not fix my issue
- [x] I have checked the [troubleshooting guide](https://isso-comments.de/docs/guides/troubleshooting/)
- [x] I have searched the [open issues](https://github.com/isso-comments/isso/issues), but my issue has not already been reported

## What is not working?

Misaka is no longer being maintained (nor is hoedown, its underlying C-based implementation) and also causes some installation headaches on some platforms. It would be beneficial to move to a pure-Python (and actively-maintained) Markdown processor.

I've been experimenting with [mistune](https://pypi.org/project/mistune/) for my newer projects and it presents a very similar API to misaka, so it shouldn't be particularly difficult for isso to switch to this.

jelmer:
Yeah, mistune looks like the most reasonable alternative today. Are you volunteering to do the migration?

cc @ix5 in case they have any objections.

fluffy-critter:
I'm not volunteering, just suggesting. I don't have a lot of energy for working on software these days.

PR #1079, opened 2025-12-22
Replace the Markdown parser and HTML renderer Misaka with Mistune #1040

ritzmann:
Misaka has been abandoned several years ago. Misaka depends on Hoedown,
a Markdown parser implemented in C, which limits platform support.

Mistune provides a similar API and features, is a pure Python implementation and is actively
maintained.

The goal is to maintain backwards compatibility with the existing markdown options where
feasible and render HTML that closely resembles the implementation with Misaka.

Syntax differences
==================

HTML elements
-------------

Misaka allows HTML expressions like ``<em>Hi</em>`` in Markdown. This expression
will be rendered to HTML as is: ``<em>Hi</em>``. Mistune takes a stricter approach
and replaces any HTML element with rendered text like this: ``&lt;em&gt;Hi&lt;/em&gt;``.

Strong and emphasis
-------------------

Misaka interprets asterisks like this:

1. \*\*word\*\*: Strong **word**
2. \*word\*: Emphasized *word*
3. \*\*word1\*\*\*word2\*: Strong **word1** and emphasized *word2*

Mistune is the same on the first two points but does not interpret asterisks inside words:

3. \*\*word1\*\*\*word2\*: Renders as \*\*word1\*\*\*word2\*

If you want to have the same rendered result with Mistune, use underscores for emphasis:

3. \*\*word1\*\*\_word2\_: Strong **word1** and emphasized *word2*

Fenced code
-----------

In Misaka, fenced code can be marked by enclosing the code block in three backticks
\`\`\` or by putting four space characters in front of each code line. Mistune
only understands three backticks.

Plugins
=======

Fenced-code
-----------

Mistune always renders fenced code. The behavior is not configurable and no
fenced-code plugin exists for Mistune.

Superscript
-----------

This plugin exists in both engines with the same name ``superscript``. The
Markdown syntax is different however.

Misaka: ``^(superscripted_text)``

Mistune: ``^superscripted_text^``

### DIFF
--- a/isso/html/__init__.py
+++ b/isso/html/__init__.py
@@ -74,8 +74,7 @@ class Markup(object):
             from isso.html.mistune import MistuneMarkdown
             self.parser = MistuneMarkdown(conf.section('markup.mistune'))
             logging.info("Using Mistune as Markdown rendering engine")
-        elif ((not conf_markup.has_option('renderer')) or
-              (conf_markup.has_option('renderer') and conf_markup.get('renderer') == "misaka")):
+        elif not conf_markup.has_option('renderer') or conf_markup.get('renderer') == "misaka":
             # We do not want to depend on Misaka unless it is actually used
             from isso.html.misaka import MisakaMarkdown
             self.parser = MisakaMarkdown(conf)


isso/utils/html.py
```diff
@@ -96,17 +95,27 @@ def blockcode(self, text, lang):
 
 class Markup(object):
 
-    def __init__(self, conf):
-        self.flags = conf.getlist("flags")
-        self.extensions = conf.getlist("options")
+    # Map the name of each misaka extension to a corresponding mistune plugin.
+    # The mistune plugin may be None if mistune does not support similar functionality or it is already built in.
```
jelmer: This seems sketchy to me - can you at least add plenty of tests for it?

jelmer: Also, can you list the types of constructs we're losing support for in the PR description?

jelmer (changes requested):
Thanks! This looks reasonable to me overall. I think the main doubts are around automatically attempting to rewrite extensions.

isso/utils/html.py
ritzmann: Yes, I am planning to add a test for each misaka extension and get as close as possible to its rendering.

ritzmann: Actually, I was thinking about adding plugins or customize mistune to get the same set of options. Some things might be done out of the box by mistune already, needs testing.

ritzmann:
@jelmer, I am currently looking into what it would take to make a clean break and I am wondering what to do about the server config:
```INI
[markup]
options = strikethrough, superscript, autolink, fenced-code
flags =
allowed-elements =
allowed-attributes =
```

The above has the issue that `options` really means Misaka extensions, not some generic set of markup options. The same goes for `flags`. Would it make sense to use a nested section in the configuration file that takes Mistune-specific settings like this:
```INI
[markup]
allowed-elements =
allowed-attributes =

[markup.mistune]
plugins = strikethrough, superscript, subscript
arguments = escape, hard_wrap
```

jelmer:
That looks reasonable. It might make sense to support both mistune and misaka for a while (with sections named as you're suggesting), then warn people that misaka is going away and then eventually remove it.

ritzmann:
DONE:
* Implemented code to separate Misaka from Mistune configuration.
* Instantiate Misaka or Mistune renderer depending on configuration.

TODO:
* Undo removal of Misaka pieces
* Add unit tests for Mistune rendering
* More logging
* Find where Mistune is rendering Markdown differently and document
* Update documentation

I wrote earlier:
> Look into support Python smartypants

I am confused where that came from. Doesn't seem to make much sense or at least not as part of this change.

docs/docs/reference/server-config.rst
```diff
@@ -390,7 +390,7 @@ require-email
 Markup
 ------
 
-Customize markup and sanitized HTML. Currently, only Markdown (via Misaka) is
+Customize markup and sanitized HTML. Currently, only Markdown (via mistune) is
```
jelmer: This needs updating

docs/docs/technical-docs/server.rst
```diff
@@ -8,7 +8,7 @@ Isso uses some of the following dependencies:
 
 - `werkzeug <https://github.com/mitsuhiko/werkzeug>`_ – WSGI toolkit
 - `itsdangerous <https://github.com/mitsuhiko/itsdangerous>`_ – store signed data on untrusted clients
-- `misaka <http://misaka.61924.nl/>`_ – fast Markdown processor written in C
+- `mistune <https://mistune.lepture.com/>`_ – a fast yet powerful Python Markdown parser
```
jelmer: Same, probably mentioning one or the other is required, and that misaka is deprecated

ritzmann: Noted

isso/utils/html.py
ritzmann: Removed mapping completely

ritzmann:
One minor item to be resolved still:
https://github.com/isso-comments/isso/pull/1079#discussion_r2646349557

And I still need to do more manual testing but otherwise this is ready from my point of view.

PR #1104, opened 2026-04-04
Add per-thread read-only mode

### LINKED ISSUE #1039: Mechanism to display thread only (no reply box)
## Checklist
- [x] I am running the latest version. Installing Isso from GitHub from the `master` branch does not fix my issue
- [x] I have searched the [open issues](https://github.com/isso-comments/isso/issues), but my issue has not already been reported

This is a feature request, for the ability to have a read-only view of a comment thread, where it would render the thread out but not provide any reply functionality. This could possibly be done as a `data-` configuration flag on the `<script src>`, and would allow implementing "soft" closed comments as well as some other uses, such as requiring login to the site itself to be able to post comments.

Obviously this wouldn't prevent someone who is truly dedicated to posting to a "locked" thread from doing so, but it would at least put in place a barrier that most users would not be able to cross.

pkvach:
## Checklist
      

A thread can be marked read-only from the new Threads page of the admin interface (/admin/threads/). The state is stored per-thread in the database and enforced on the server: the comment-fetch response reports it, and new comments, edits and deletions are rejected for a read-only thread. Existing comments and votes are unaffected.

The embed client reads the flag from the fetch response and hides the postbox and the reply/edit/delete controls to match.

<details><summary>Example</summary>
<p>

- Normal mode:

image

- Read-only mode:

image

</p>
</details> 

This feature is useful for archived content, closed discussions, or when comments should be visible but not interactive.

## Why is this necessary?

Related to: #1039

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,6 +1,17 @@
 Changelog for Isso
 ==================
 
+%(version)s (%(date)s)
+--------------------
+
+New Features
+^^^^^^^^^^^^
+
+- Client: Add ``data-isso-read-only`` client option to render threads in
+  read-only mode (hides postbox and disables reply/edit/delete UI). (`#1104`_, pkvach)
+
+.. _#1104: https://github.com/isso-comments/isso/pull/1104
+
 0.14.0 (2026-03-26)
 --------------------
 
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -8,8 +8,11 @@ New Features
 ^^^^^^^^^^^^
 
 - Add option to show/hide website field in comment form (`#1111`_, pkvach)
+- Client: Add ``data-isso-read-only`` client option to render threads in
+  read-only mode (hides postbox and disables reply/edit/delete UI). (`#1104`_, pkvach)
 
 .. _#1111: https://github.com/isso-comments/isso/pull/1111
+.. _#1104: https://github.com/isso-comments/isso/pull/1104
 
 0.14.0 (2026-03-26)
 --------------------
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -12,12 +12,17 @@ New Features
 - Add no-cache headers to fix stale comment display (`#1110`_, pkvach)
 - Add webhook support for comment events (`#1123`_, p1slave)
 - Update comment counts in the admin interface dynamically after moderation. (`#1113`_, pkvach)
+- Read-only comment threads: per-thread state stored in the database, toggled
+  from a new *Threads* page in the admin interface, enforced on
+  ``new``/``edit``/``delete``. The embed client hides the postbox and
+  the reply/edit/delete UI for a read-only thread. (`#1104`_, pkvach)
 
 .. _#1111: https://github.com/isso-comments/isso/pull/1111
 .. _#1117: https://github.com/isso-comments/isso/pull/1117
 .. _#1110: https://github.com/isso-comments/isso/pull/1110
 .. _#1123: https://github.com/isso-comments/isso/pull/1123
 .. _#1113: https://github.com/isso-comments/isso/pull/1113
+.. _#1104: https://github.com/isso-comments/isso/pull/1104
 
 0.14.0 (2026-03-26)
 --------------------
--- a/docs/docs/reference/client-config.rst
+++ b/docs/docs/reference/client-config.rst
@@ -272,6 +272,30 @@ data-isso-sorting
 
     .. versionadded:: 0.13.1
 
+.. _data-isso-read-only:
+
+data-isso-read-only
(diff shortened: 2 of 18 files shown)


jelmer (changes requested):
Any reason this is client side only? If we stored this on the server than we could: actually enforce this, allow controlling it from e.g. the admin panel, wouldn't need JavaScript hackery to enable.

pkvach:
> Any reason this is client side only? If we stored this on the server than we could: actually enforce this, allow controlling it from e.g. the admin panel, wouldn't need JavaScript hackery to enable.

Thanks for the feedback. 
You are right. Server-side enforcement gives stronger control and real locking.

I chose a small and safe change first. A client-side flag is quick to add and handles simple cases like soft close or login requirement. No backend or database changes.

A server approach needs more work, thing like:

- Add thread read and update APIs.
- Update database schema.
- Build admin UI to manage thread state.
- Make the client check thread status on each load.

I avoided this scope for now, though I agree it makes sense for a future update.

If there's interest in a server-side solution, I'd be happy to help plan or contribute to that as a follow-up.

BBaoVanC:
I think this only makes sense to merge if it were server side. Otherwise someone could trivially edit or bypass the client-side UI and post to a thread you intended to be read only. Instead of configuring read only status client-side on the page, the client could request the read only status of the thread from the server while it loads, then hide the postbox based on what the server says.

pkvach:
Yes, I understand that a server-side solution is the better approach here. 
I'll convert this PR to draft and try to rework it. Thanks for your feedback.

fluffy-critter:
I think that this client-side-only approach would handle my particular use case pretty well and I'd still like to see it show up. The purpose of the original feature request was to make posting ability conditional based on aspects of the client and it's not like Isso conversations are a high-security thing; having something that's a basic barrier to entry for AI agents and drive-by trolls is all I particularly care about.

Also, a situation where the only way to close comments on a thread is for an admin to shut it off entirely within the database is completely useless for my needs.

An approach where the server presents some sort of signed token to the client (using a basic HMAC-type approach), which needs to be present for the post to work, would be nice. For example, having `data-comment-token="[some random string]|[md5sum of the same random string concatenated with a secret security key]"`, and server-side configuration where if the security key is present and data-comment-token isn't, the reply field doesn't appear. That also wouldn't require any database changes on isso, and would satisfy all of the use cases I can think of.

jelmer:
> I think that this client-side-only approach would handle my particular use case pretty well and I'd still like to see it show up. The purpose of the original feature request was to make posting ability conditional based on aspects of the client and it's not like Isso conversations are a high-security thing; having something that's a basic barrier to entry for AI agents and drive-by trolls is all I particularly care about.
> 
> Also, a situation where the only way to close comments on a thread is for an admin to shut it off entirely within the database is completely useless for my needs.
> 
> An approach where the server presents some sort of signed token to the client (using a basic HMAC-type approach), which needs to be present for the post to work, would be nice. For example, having `data-comment-token="[some random string]|[md5sum of the same random string concatenated with a secret security key]"`, and server-side configuration where if the security key is present and data-comment-token isn't, the reply field doesn't appear. That also wouldn't require any database changes on isso, and would satisfy all of the use cases I can think of.

Why wouldn't isso comments be a high security thing? I don't see why they couldn't be - and at the very least we shouldn't do things to make it's security profile worse.

Also, wouldn't users still be able to edit their comments with this PR despite a thread being read only? And users can still delete their comments? And pretty sure AIs will easily work out how to just send the magic POST request necessary.

I think if this was called something like "hide comment create form" I could be onboard with it. As implemented, I think it is misleading users into thinking it is more than it is.

fluffy-critter:
Fair points! I think the main thing I was getting at is that the proposed alternative of doing purely server-side thread-level locking isn't suitable for the original request.

Calling this PR "hide comment create form" would be fine. I'll update the original issue with my implementation thoughts, in any case.

fluffy-critter:
@pkvach Being able to lock a thread from the admin panel is useful but it's not what I was looking for with this; for example, I was hoping to make it so that I can require someone to log in to my website to be able to post a comment, while still being able to display the thread to people who aren't logged in. As such, this PR wouldn't actually close the referenced issue.

pkvach:
> @pkvach Being able to lock a thread from the admin panel is useful but it's not what I was looking for with this; for example, I was hoping to make it so that I can require someone to log in to my website to be able to post a comment, while still being able to display the thread to people who aren't logged in. As such, this PR wouldn't actually close the referenced issue.

Thanks for the feedback! I agree that the current changes don't meet your needs, so I'll change the reference to the issue. I may consider working on that implementation once this PR is resolved.

fluffy-critter:
Excellent, thanks for understanding. :)

PR #1107, opened 2026-04-20
docs: mistune: Remove incorrect claim about fenced code blocks

vincentbernat:
The guide claimed Mistune only understands three backticks, but it actually supports 4-space indented code blocks too.

    import mistune
    md = mistune.create_markdown(escape=True)
    md('    indented code block\n    second line\n')
    md('```\nbacktick fenced\n```\n')

Both render as `<pre><code>...</code></pre>`.

## Checklist
- [X] All new and existing **tests are passing**
- [X] I have written **proper commit message(s)**

### DIFF
--- a/docs/docs/guides/mistune.rst
+++ b/docs/docs/guides/mistune.rst
@@ -47,13 +47,6 @@ If you want to have the same rendered result with Mistune, use underscores for e
 
 3. \*\*word1\*\*\_word2\_: Strong **word1** and emphasized *word2*
 
-Fenced code
-^^^^^^^^^^^
-
-In Misaka, fenced code can be marked by enclosing the code block in three backticks
-``````` or by putting four space characters in front of each code line. Mistune
-only understands three backticks.
-
 Plugins
 -------
 


docs/docs/guides/mistune.rst
```diff
@@ -47,13 +47,6 @@ If you want to have the same rendered result with Mistune, use underscores for e
 
 3. \*\*word1\*\*\_word2\_: Strong **word1** and emphasized *word2*
 
-Fenced code
-^^^^^^^^^^^
-
-In Misaka, fenced code can be marked by enclosing the code block in three backticks
```
jelmer, line 53: This says Misaka rather than Mistune.

vincentbernat: Yes. It says Misaka can understand both and Mistune only backticks. This is incorrect.

jelmer (approved):

docs/docs/guides/mistune.rst
jelmer: Ah, sorry - I missed that last bit. Thanks!

PR #1110, opened 2026-04-21
views: Add no-cache headers to fix stale comment display

### LINKED ISSUE #939: Missing comments (but count is correct!)
## Checklist
- [ ] I am running the latest version. Installing Isso from GitHub from the `master` branch does not fix my issue
- [x] I have searched the [open issues](https://github.com/posativ/isso/issues), but my issue has not already been reported

## What is not working?

Not all comments are loaded. The comment count on the top of the page is always correct, but at the bottom of the article not all comments are shown. I noticed that sometimes switching from `http` to `https` in the URL bar of the browser helps, but not always. See for example:
- http://www.mardy.it/blog/2022/12/leaving-canonical-again.html
- https://www.mardy.it/blog/2022/12/leaving-canonical-again.html

I haven't been able to figure out a reproducible pattern.

I'm running isso 0.13.0 as fcgi, and this is my `host` stanza from the config file:

```
; your website or blog (not the location of Isso!)
host =
  http://www.mardy.it/
  https://www.mardy.it/
  http://mardy.it/
  https://mardy.it/
  http://blog.mardy.it/
  https://blog.mardy.it/
```
I had the same issue with isso 0.12.x, so it's not a regression.

pkvach:
## Checklist
      

Add no-cache HTTP header on dynamic JSON GET endpoints for comment listings and views. This prevents browsers and proxies from caching stale comment data, ensuring users always see the latest comments.

Fixes issue where new comments were missing from the UI unless the browser cache was disabled. The comment count was always correct, but displayed comments could be outdated due to caching.

## Why is this necessary?

Fixes https://github.com/isso-comments/isso/issues/939

### DIFF
--- a/isso/utils/__init__.py
+++ b/isso/utils/__init__.py
@@ -13,6 +13,13 @@ from werkzeug.exceptions import BadRequest
 from isso.wsgi import Request
 
 
+NO_CACHE_HEADERS = {
+    "Cache-Control": "no-cache, no-store, must-revalidate",
+    "Pragma": "no-cache",
+    "Expires": "0",
+}
+
+
 def anonymize(remote_addr):
     """
     Anonymize IPv4 and IPv6 :param remote_addr: to /24 (zero'd)
@@ -32,6 +39,21 @@ def anonymize(remote_addr):
             return "" + ipv6.exploded.rsplit(":", 5)[0] + ":" + ":".join(["0000"] * 5)
         except ipaddress.AddressValueError:
             return "0.0.0.0"
+
+
+def set_no_cache_headers(resp):
+    """
+    Set HTTP headers to prevent caching of dynamic content.
+
+    Args:
+        resp: A Response object to add headers to
+
+    Returns:
+        The same Response object with no-cache headers added
+    """
+    for key, value in NO_CACHE_HEADERS.items():
+        resp.headers[key] = value
+    return resp
 
 
 class Bloomfilter:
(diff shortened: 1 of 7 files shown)


jelmer:
Can we just unconditionally set nocache? Why does this need to be another configuration option?

isso/utils/__init__.py
```diff
@@ -13,6 +13,13 @@
 from isso.wsgi import Request
 
 
+NO_CACHE_HEADERS = {
+    "Cache-Control": "no-cache, no-store, must-revalidate",
+    "Pragma": "no-cache",
```
jelmer: This is a request header. Why is it here?

isso/utils/__init__.py
```diff
@@ -13,6 +13,13 @@
...
+    "Expires": "0",
```
jelmer: is only a fallback for HTTP/1.0 caches, and is redundant whenever Cache-Control is present - Cache-Control takes precedence.

isso/utils/__init__.py
```diff
@@ -13,6 +13,13 @@
 from isso.wsgi import Request
 
 
+NO_CACHE_HEADERS = {
+    "Cache-Control": "no-cache, no-store, must-revalidate",
```
jelmer: no-store + must-revalidate are redundant alongside each other: no-store already forbids storing, so there's nothing left to revalidate.

pkvach:
Good point. I agree that making this a config option might be overkill for this case. Thanks for the feedback!

jelmer (changes requested):

jelmer (approved):

Issue #501, opened 2018-11-11
admin panel doesn't refresh in some cases


neocogent:
I know you're working on this right now so in case it wasn't noticed I just thought I'd mention a couple odd behaviors I saw. I guess it's work in progress.

After I click a "Activate"/"Delete" moderation button the list doesn't refresh for me. And if "group by thread" is not checked then the Valid/Pending selection doesn't show correct items. Once I marked the group by thread it seems list shows expected items.

Also, the Isso Admin title links to a non-existing url. Maybe have it link back to actual site or just reload admin page. It would be nice to have a link top-right, perhaps, that links to actual site. And a logout button. I'd be willing to do some of this code if it's not already in the works.

PR #1113, opened 2026-05-26
js: admin: update comment counts dynamically after moderation

pkvach:
## Checklist
      

> [!NOTE]
> I've created a separate PR for fixing issue with E2E tests: https://github.com/isso-comments/isso/pull/1114

Update the "Valid", "Pending", and "Staled" count badges in the admin interface immediately after a comment is activated or deleted. This eliminates the need for a manual page reload.

## Why is this necessary?

Related #501

### DIFF
--- a/isso/js/admin.js
+++ b/isso/js/admin.js
@@ -24,12 +24,32 @@ function fade(element) {
         op -= op * 0.1;
     }, 10);
 }
+const modeCountIds = {1: 'count-mode-1', 2: 'count-mode-2', 4: 'count-mode-4'};
+
+function setCountBadge(modeId, count) {
+    const el = document.getElementById(modeId);
+    if (!el) return;
+    const countEl = el.querySelector('.count');
+    if (!countEl) return;
+    countEl.textContent = count;
+}
+
 function moderate(com_id, hash, action, isso_host_script) {
-  ajax({method: "POST",
-        url: isso_host_script + "/id/" + com_id + "/" + action + "/" + hash,
-        success: function(){
-            fade(document.getElementById("isso-" + com_id));
-        }});
+    ajax({method: "POST",
+          url: isso_host_script + "/id/" + com_id + "/" + action + "/" + hash,
+          success: function(responseText){
+              fade(document.getElementById("isso-" + com_id));
+              try {
+                  const data = JSON.parse(responseText);
+                  if (data.counts) {
+                      for (const [mode, modeId] of Object.entries(modeCountIds)) {
+                          setCountBadge(modeId, data.counts[mode] || 0);
+                      }
+                  }
+              } catch (e) {
+                  console.error("Failed to parse moderate response:", e);
+              }
+          }});
 }
 function edit(com_id, hash, author, email, website, comment, isso_host_script) {
   ajax({method: "POST",
@@ -133,4 +153,9 @@ function log_out() {
 function toggleTooltip(tooltipContainer) {
     const tooltipText = tooltipContainer.querySelector(".search-tooltip-text");
     tooltipText.classList.toggle("show");
+}
+
+// Guard needed: admin.js is served raw (not bundled), so module is undefined in browsers
+if (typeof module !== 'undefined') {
+    module.exports = { setCountBadge };
 }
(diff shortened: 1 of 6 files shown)


jelmer (changes requested):
Thanks for this - the approach is right and the diff is pleasantly small. Wrapping the numbers in a dedicated <span class="count"> and stamping data-mode on the comment div is the correct seam, and it's nice to see admin.js get its first unit tests.

Two correctness problems in the delete/activate path, though, and both come from the client trying to compute deltas it can't actually derive. First, Comments.delete() has two outcomes: if nothing references the comment as a parent the row really goes away, but if it is referenced the row survives as a mode=4 tombstone. The new code assumes the first case unconditionally, so deleting a replied-to Pending comment renders Pending -1 with Staled unchanged, when the truth is Pending -1 / Staled +1. Deleting an already-staled but referenced comment decrements Staled against a database that didn't change, and _remove_stale() can cascade-delete an arbitrary number of further rows in the same request. Second,  the moderate view returns "Already activated" with status 200 when item["mode"] == 1, and ajax() treats any 200 as success — so activating twice (or following an activate link from a notification email for a comment already approved in the UI) moves Valid +1 and the source badge -1 with no underlying change. Both of these leave the badges wrong until a manual reload, which is the thing the PR set out to eliminate.

I'd suggest fixing this server-side rather than patching the arithmetic: have the activate and delete branches of moderate return the fresh count_modes() dict as JSON, and have the JS set the badges from the response. That's a smaller change than it sounds, it resolves both bugs at once, and it's robust against the _remove_stale() cascade that no client-side  calculation can model. Two minor things while you're in there: admin.js is served raw via <script src> rather  than bundled like embed.js/count.js, so the new module.exports ships to the browser — the typeof module !== 'undefined' guard makes that safe, but a one-line comment explaining why it's there would help the next reader.

pkvach:
Thanks for the great and detailed feedback, @jelmer.
The server-side approach makes a lot of sense. I'll get to the updates in the next few days.

isso/views/comments.py
```diff
@@ -847,7 +847,7 @@ def moderate(self, environ, request, id, action, key):
             with self.isso.lock:
                 self.comments.activate(id)
             self.signal("comments.activate", thread, item)
-            return Response("Comment has been activated", 200)
+            return JSON({"counts": self.comments.count_modes()}, 200)
```
jelmer, line 853: "Already activated" above should also return JSON

isso/js/admin.js
```diff
@@ -24,12 +24,30 @@ function fade(element) {
...
+          success: function(responseText){
+              fade(document.getElementById("isso-" + com_id));
+              try {
+                  const data = JSON.parse(responseText);
+                  if (data.counts) {
+                      for (const [mode, modeId] of Object.entries(modeCountIds)) {
+                          setCountBadge(modeId, data.counts[mode] || 0);
+                      }
+                  }
+              } catch (e) {}
```
jelmer: catch (e) {} violates project guidelines

isso/views/comments.py
```diff
@@ -802,11 +802,11 @@ def unsubscribe(self, environ, request, id, email, key):
                     }
                 </script>
 
-    @apiSuccessExample Delete using POST:
-        Comment has been deleted
+    @apiSuccessExample {json} Delete using POST:
+        {"counts": {"1": 3, "2": 0, "4": 1}}
```
jelmer: count_modes() returns integer keys

isso/js/tests/unit/admin.test.js
```diff
@@ -0,0 +1,34 @@
+/**
```
jelmer, line 1: Does jest actually pick up this file?

isso/views/comments.py
pkvach: Yes, that's true, but JSON converts integer keys to strings. I added assertions to the tests to verify this behavior.

isso/js/tests/unit/admin.test.js
pkvach: Yes, Jest picks up the file. All tests in the directory are run.

Here is a link to a GitHub Actions run to show that:
https://github.com/isso-comments/isso/actions/runs/32827183814/job/97737667757?pr=1113

isso/views/comments.py
jelmer: for consistency, it would be good to reflect that in the docstrings - other consumers might be more strict

jelmer (approved):
LGTM

PR #1117, opened 2026-07-29
Harden `API.verify`

NicolaiSoeborg:
No key should be larger than 2**16. Currently I believe it is possible to "spam" the "Author" field with very large keys

## Checklist
      

Maximum field size for comments (new/edit/etc) is 2**16

## Why is this necessary?

I was looking though the code and saw that "author" field wasn't covered by the current max size check. While the current code is a bit redundant I think that is best to cover all current and future cases.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -8,8 +8,10 @@ New Features
 ^^^^^^^^^^^^
 
 - Add option to show/hide website field in comment form (`#1111`_, pkvach)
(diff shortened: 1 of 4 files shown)


jelmer (dismissed):
Please fix the test failures, as well as addressing the check list items in the PR description

jelmer:
Please see items in the PR description, specifically:

* Update documentation to mention these limits
* Update commit messages
* Add entry to CHANGES.rst

See also https://isso-comments.de/docs/contributing/

PR #1127, opened 2026-09-06
migrate: Add --thread-id to key Disqus threads by identifier

### LINKED ISSUE #563: Disqus import not keeping the thread IDs
Hi, I'm trying to migrate my site from Disqus to isso, and while I was able to import my comments and the importer showed the specified thread IDs during the import, the actual database is still using the page URI as the thread identifier. For example:

```
sqlite> select uri,title from threads limit 5;
/comics/journal/400-Shoes|Journal: Shoes
/comics/journal/278-Things-that-go-splat-in-the-night|Journal: Things that go splat in the night
/comics/journal/322-Explosion|Journal: Explosion
/comics/journal/353-Rejection|Journal: Rejection
/comics/journal/469-Scary-pizza|Journal: Scary pizza
```

IN the above example, the thread with a thread ID of `/comics/journal/400-Shoes` should actually have a thread ID of `mt_1949`. When my site is configured to pull the comments with e.g.

```
<div id="isso-thread" data-thread-id="mt_1949">
```

the script correctly tries to retrieve the data with `OPTIONS /?uri=mt_1949` but that doesn't return any results, as expected.

So, is there any way to tell the Disqus importer to use the specified thread ID rather than the URI?

pkvach:
## Checklist
      

The Disqus importer keys every thread by its page URL path. If a Disqus site sets a custom `disqus_identifier` and the pages use `data-isso-id`, the imported thread lands under the URL path instead of that identifier, so the page renders with zero imported comments.

**Changes.**

- Add an opt-in `--thread-id` flag to `isso import -t disqus`. With it, a thread that has a non-empty `<id>` (the `disqus_identifier`) is keyed by that identifier; threads without one fall back to the URL path. Default import behaviour is unchanged, so existing installs are not re-keyed on re-import.
- Fix empty-`<id>` detection. The old check relied on the truthiness of an `ElementTree` element, which is `False` for an element with no subelements regardless of its text ([Element docs](https://docs.python.org/3/library/xml.etree.elementtree.html#element-objects)), so a text-only `<id>some-id</id>` was always treated as empty and the thread was skipped unless `--empty-id` was passed. It now reads the element's text.

**How to try it.**

```console
isso -c isso.cfg import -t disqus --thread-id disqus.xml
```

Then load a page whose data-isso-id matches a disqus_identifier from the export and confirm the migrated comments show up.

## Why is this necessary?

Fixes #563


isso/__init__.py
```diff
@@ -250,6 +250,12 @@ def main():
     imprt.add_argument(
         "--empty-id", dest="empty_id", action="store_true", help="workaround for weird Disqus XML exports, #135"
     )
+    imprt.add_argument(
+        "--thread-id",
+        dest="use_thread_id",
+        action="store_true",
+        help="Disqus only: key threads by disqus_identifier (<id>) instead of the page URL, #563",
```
jelmer: please don't add bug number - it's irrelevant to users

isso/__init__.py
```diff
@@ -280,7 +286,7 @@ def main():
             dbpath = conf.get("general", "dbpath")
 
         mydb = db.SQLite3(dbpath, conf)
-        migrate.dispatch(args.type, mydb, args.dump, args.empty_id)
+        migrate.dispatch(args.type, mydb, args.dump, args.empty_id, args.use_thread_id)
```
jelmer: please switch to kewyword arguments

CHANGES.rst
```diff
@@ -19,6 +19,17 @@ New Features
 .. _#1123: https://github.com/isso-comments/isso/pull/1123
 .. _#1113: https://github.com/isso-comments/isso/pull/1113
 
+Bugfixes & Improvements
+^^^^^^^^^^^^^^^^^^^^^^^
+
+- Disqus migration: add ``--thread-id`` to key imported threads by the
```
jelmer: please keep these entries short, consistent with style elsewhere

jelmer (changes requested):
LGTM, some minor comments inline

jelmer (approved):

